PR #2, opened 2015-06-11
Take data providers into account when counting tests to be executed

losipiuk:
Previously we counted method with data providers as single test
execution. Therefore all methods count was off and logs from
ProgressLoggingListener looked weird in such case.

Now we execute data provider within RequirementsExpanderInterceptor
which does the counting - and increment the seenMethodsCount by number
of paremeters sets.

### DIFF
--- a/tempto-core/src/main/java/com/teradata/tempto/internal/DataProviders.java
+++ b/tempto-core/src/main/java/com/teradata/tempto/internal/DataProviders.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.teradata.tempto.internal;
+
+import org.testng.ITestNGMethod;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import java.lang.reflect.Method;
+import java.util.Optional;
+
+import static java.util.Arrays.asList;
+
+/**
+ * Helper routines for working with TestNG data providers.
+ */
+public class DataProviders
+{
+
+    /**
+     * Obtains list of parameter sets for test method based on defined data provider.
+     * If no data provider is defined then {@link Optional#empty()} is returned.
+     * This method takes into consideration only data providers defined through annotation.
+     */
+    public static Optional<Object[][]> getParametersForMethod(ITestNGMethod method)
+    {
+        Test testAnnotation = method.getConstructorOrMethod().getMethod().getAnnotation(Test.class);
+        Class dataProviderClass = testAnnotation.dataProviderClass();
+        if (dataProviderClass == null || dataProviderClass == Object.class) {
+            dataProviderClass = method.getRealClass();
+        }
+        String dataProviderName = testAnnotation.dataProvider();
+        if (dataProviderName.isEmpty()) {
+            return Optional.empty();
+        }
+
+        Optional<Method> dataProviderMethodOptional = asList(dataProviderClass.getMethods()).stream().filter(
+                m -> {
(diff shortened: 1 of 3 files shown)


losipiuk:
@arhimondr, @sopel39, @kokosing Take a look at this in some free time.

tempto-core/src/main/java/com/teradata/tempto/internal/initialization/RequirementsExpanderInterceptor.java
```diff
@@ -80,6 +88,49 @@ public static int getMethodsCountFromContext(ITestContext context)
...
+            Optional<Object[][]> parametersForMethod = getParametersForMethod(newExpandedMethod);
+            if (parametersForMethod.isPresent()) {
+                seenMethodsCount += parametersForMethod.get().length;
+            } else {
+                seenMethodsCount++;
+            }
+        }
+    }
+
+    private Optional<Object[][]> getParametersForMethod(IMethodInstance method)
```
kokosing: Maybe put code responsible for data providers (this reflection code) into separate class and add proper testing.

tempto-core/src/main/java/com/teradata/tempto/internal/initialization/RequirementsExpanderInterceptor.java
```diff
@@ -80,6 +88,49 @@ public static int getMethodsCountFromContext(ITestContext context)
...
+        if (dataProviderClass == null || dataProviderClass == Object.class) {
+            dataProviderClass = method.getMethod().getRealClass();
+        }
+        String dataProviderName = testAnnotation.dataProvider();
+        if (dataProviderName.isEmpty()) {
+            return Optional.empty();
+        }
+
+        Optional<Method> dataProviderMethod = asList(dataProviderClass.getMethods()).stream().filter(
+                m -> {
```
kokosing: Do you need to scan all the methods? I thought DataProvider is an interface, where only one method returns an array of parameters.

tempto-core/src/main/java/com/teradata/tempto/internal/initialization/RequirementsExpanderInterceptor.java
```diff
@@ -80,6 +88,49 @@ public static int getMethodsCountFromContext(ITestContext context)
...
+                    DataProvider annotation = m.getAnnotation(DataProvider.class);
+                    return annotation != null && annotation.name().equals(dataProviderName);
+                }
+        ).findFirst();
+        if (dataProviderMethod.isPresent()) {
+            try {
+                return Optional.of((Object[][]) dataProviderMethod.get().invoke(method.getInstance()));
```
kokosing: ah... it is a static method, so what if there are multiple data provider static methods, will you just find first of them?

kokosing:
I think that using reflection is errorprone, so unit testing for that is needed. Have you check testng code, maybe they have already something to retrieve test parameters for test?

losipiuk:
@kokosing I looked into testNG code. They have sth but it is very convoluted. Have more responsibilities than just listing parameterSets based on data providers - and requires passing some weird dependencies I do not have here.

tempto-core/src/main/java/com/teradata/tempto/internal/initialization/RequirementsExpanderInterceptor.java
losipiuk: DataProvider is an annotation you can add to the method. It has a name and then data provider can be referenced in test method using \@Test(dataProvider = "dataProviderName").
So we have to scan check all methods with \@DataProvider annotation and pick the one with name matching one used by specific test method.

tempto-core/src/main/java/com/teradata/tempto/internal/initialization/RequirementsExpanderInterceptor.java
losipiuk: There should be just one

losipiuk:
is that ok now?

tempto-core/src/main/java/com/teradata/tempto/internal/DataProviders.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+public class DataProviders
+{
+
+    /**
+     * Obtains list of parameter sets for test method based on defined data provider.
+     * If no data provider is defined then {@link Optional#empty()} is returned.
+     * This method takes into consideration only data providers defined through annotation.
+     */
+    public static Optional<Object[][]> getParametersForMethod(ITestNGMethod method)
```
kokosing: static?

kokosing:
looks good

losipiuk:
merged

PR #12, opened 2015-06-23
Support statement timeout

brian-rickman:
The primary motivation for this change is to support a new automated test that will make use of a random sql query generator.  Sometimes the generator produces queries which may run for an excessively long time.  If this occurs, it is preferable to have the offending query canceled after a few minutes so that other queries can be executed.

Added ability to specify statement timeout in database configuration file.  Refactored JdbcConnection by removing unused constructors.  Cleaned up readme and sample configuration file.  Insert the version number into the wrapper script as part of the build so that the correct jar is loaded at runtime.

Tested with a variety of queries and timeout settings to prove that queries which exceed the specified timeout are canceled, and that queries which take less time are allowed to run to completion.


expected-result-generator/bin/generate_results.py
```diff
@@ -10,8 +10,11 @@
 import argparse
 import subprocess
 import os
+import glob
 
-CLASSPATH = ".:../build/libs/expected-result-generator-all.jar"
+jar_file_list = glob.glob("../build/libs/expected-result-generator-all*.jar")
```
kokosing: this probably does not belong to this change... Please mention about that in commit message

expected-result-generator/src/main/java/com/teradata/tempto/sql/JdbcConnection.java
```diff
@@ -75,44 +64,22 @@ public JdbcConnection(String propertiesFileName) throws IOException {
...
-  }
-
-  /**
-   * @param jdbcUrl    jdbc connection URL
-   * @param user       Database user name
-   * @param password   Database password
-   */
-  public JdbcConnection(String jdbcUrl, String user, String password) {
-    initialize(jdbcUrl, user, password);
+    LOGGER.info("timeout {}", timeout);
```
kokosing: join this 4 log statements into one

expected-result-generator/src/main/java/com/teradata/tempto/sql/JdbcConnection.java
```diff
@@ -75,44 +64,22 @@ public JdbcConnection(String propertiesFileName) throws IOException {
...
-    String url = properties.getProperty(JDBC_URL);
-    String user = properties.getProperty(JDBC_USER);
-    String password = properties.getProperty(JDBC_PASSWORD);
-
-    initialize(url, user, password);
-  }
+    jdbcUrl = properties.getProperty(JDBC_URL);
+    user = properties.getProperty(JDBC_USER);
+    password = properties.getProperty(JDBC_PASSWORD);
+    timeout = Integer.parseInt(properties.getProperty(JDBC_TIMEOUT, "600"));
```
kokosing: document that 10 minutes is a default timeout

expected-result-generator/src/main/java/com/teradata/tempto/sql/SqlResultGenerator.java
```diff
@@ -80,10 +67,12 @@ public static void main(String[] args) {
       resultGenerator.generateExpectedResults();
     } catch (ParseException e) {
       usage(options);
+      System.exit(99);
```
kokosing: do not use System.exit()... ever

expected-result-generator/src/main/java/com/teradata/tempto/sql/SqlResultGenerator.java
```diff
@@ -80,10 +67,12 @@ public static void main(String[] args) {
...
     } catch (Exception e) {
       LOGGER.error("Caught exception in main", e);
+      System.exit(99);
```
kokosing: what does 99 mean?

kokosing:
Please write in the description to what component this change is related to. Also write about what steps you took to make sure it is well tested.

expected-result-generator/bin/generate_results.py
brian-rickman: Updated description to show that this change is intentional.

losipiuk: You should not use * here.  You can easily end up with multiple versions of
jar in the build directory and then you will have messed up classpath.
Generally it would be best to have build target which builds distribution
package with both running script and proper jar file.

expected-result-generator/src/main/java/com/teradata/tempto/sql/JdbcConnection.java
brian-rickman: Done

expected-result-generator/src/main/java/com/teradata/tempto/sql/JdbcConnection.java
brian-rickman: Done

expected-result-generator/src/main/java/com/teradata/tempto/sql/SqlResultGenerator.java
brian-rickman: Removed

expected-result-generator/src/main/java/com/teradata/tempto/sql/SqlResultGenerator.java
brian-rickman: Removed

brian-rickman:
@kokosing Thank you for reviewing my code.  I have addressed all of your comments in my second commit.

expected-result-generator/bin/generate_results.py
brian-rickman: That does sound like a better technique.  I will look into it.

kokosing:
Thank you. Code looks good to me, modulo @losipiuk comments.

PR #33, opened 2015-08-13
Add line escaping in AnnotatedFileParser

losipiuk:
Add line escaping in AnnotatedFileParser

Sometimes we need to write common content line in .result file
which would be interpreted as special line by AnnotatedFileParser.
To be able to do this this patch adds line escape char ('\') which if
present at beginning of line is removed during parsing and remaining
part of line is treated as common content line.


sogorkis:
lgtm

sopel39:
lgtm

PR #36, opened 2015-08-25
core: remove unused TestContext parameter from JdbcQueryExecutor constructor

kokosing:
core: remove unused TestContext parameter from JdbcQueryExecutor constructor

Test Plan: none

Reviewers: ksobczak, sogorkis


sogorkis:
LGTM

PR #38, opened 2015-08-28
core: add posibility to use SshClient with pem file instead of password

kokosing:
core: add posibility to use SshClient with pem file instead of password

Moreover now user can inject into test SshClientFactory so he can create
SshClient dynamically.

Note that, when you port your tempto to version 1.0.35 and you use
SssClient then you need to modify your tempto-configuration.yaml.

Test Plan: using SshClient with pem file in tests

Reviewers: ksobczak, losipiuk, arosa

### DIFF
--- a/tempto-core/src/main/java/com/teradata/tempto/internal/ssh/JschSshClientFactory.java
+++ b/tempto-core/src/main/java/com/teradata/tempto/internal/ssh/JschSshClientFactory.java
@@ -1,0 +1,72 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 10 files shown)


tempto-core/src/main/java/com/teradata/tempto/internal/ssh/JschSshClientFactory.java
```diff
@@ -0,0 +1,72 @@
...
+    private static final Logger LOGGER = getLogger(JschSshClientFactory.class);
+
+    private final JSch jSch = new JSch();
+
+    @Override
+    public SshClient create(String host, int port, String user, Optional<String> password)
+    {
+        try {
+            String passwordToString = password.isPresent() ? '/' + password.get() : "";
+            LOGGER.debug("Allocating new SSH session to: {}{}@{}:{}", user, passwordToString, host, port);
```
sopel39: is logging password a good idea?

tempto-core/src/main/java/com/teradata/tempto/internal/ssh/JschSshClientFactory.java
```diff
@@ -0,0 +1,72 @@
...
+
+    private Properties getConfig()
+    {
+        Properties config = new Properties();
+        config.put("StrictHostKeyChecking", "no");
+        return config;
+    }
+
+    @Override
+    public void addIdentity(String pathToPem)
```
sopel39: move above getConfig

sopel39:
lgtm

PR #40, opened 2015-08-31
core: update unit tests for JschSshClient

kokosing:
core: update unit tests for JschSshClient

Modify JSchSshClientTest to reflect recent changes. Add test to check
authentication with a identity key.

Test Plan: unit tests

Reviewers: ksobczak


sopel39:
lgtm

PR #41, opened 2015-08-31
core: add utility command method to SshClient

kokosing:
core: add utility command method to SshClient

For short lasting commands with simple output I added a method command.
Thanks to that clients code will get much simpliefied in such cases.

Test Plan: none

Reviewers: losipiuk, ksobczak


sopel39:
lgtm

PR #43, opened 2015-09-02
Replace commons-logging dependecy with slf4j wrapper

losipiuk:
Replace commons-logging dependecy with slf4j wrapper

Reviewers: sogorkis


sogorkis:
LGTM

PR #48, opened 2015-09-07
ssh: throw explicit exceptions on timeout and command failure

kokosing:
ssh: throw explicit exceptions on timeout and command failure

Throwing explicit exception in case of command timeout or command
execution failure. Thanks to that user is able to handle this in
more elegant way than catching the RuntimeException.

Additionally, log command output (as debug) and log commands passed via
ssh.

Test Plan: run product tests which are using ssh

Reviewers: losipiuk


sopel39:
lgtm

losipiuk:
fix comment message

PR #55, opened 2015-09-29
examples: make examples to depend on third party libs

kokosing:
examples: make examples to depend on third party libs

Test Plan: running example tests

Reviewers: losipiuk


sopel39:
lgtm

Issue #45, opened 2015-09-03
Inject components into a method annotated with @BeforeTestWithContext


kokosing:

kokosing:
Done % generic types are not yet supported

PR #57, opened 2015-09-30
core: add possibility to inject state into before and after methods

kokosing:
core: add possibility to inject state into before and after methods

Task: #45

Test Plan: unit tests and running examples

Reviewers: losipiuk

### DIFF
--- a/tempto-core/src/main/java/com/teradata/tempto/internal/ReflectionInjectorHelper.java
+++ b/tempto-core/src/main/java/com/teradata/tempto/internal/ReflectionInjectorHelper.java
@@ -1,0 +1,101 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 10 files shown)


tempto-core/src/main/java/com/teradata/tempto/internal/ReflectionInjectorHelper.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    private Object getInstanceForParameter(Injector injector, Parameter parameter, List<Annotation> parameterAnnotation)
+    {
+        TypeLiteral type = TypeLiteral.get(parameter.getParameterizedType());
+        Optional<Annotation> namedAnnotation = parameterAnnotation.stream()
+                .filter(namedAnnotation())
+                .findFirst();
+
+        if (namedAnnotation.isPresent()) {
```
losipiuk: `@Named` is just one of possibilities. Any annotation can be part of injection key.

tempto-core/src/main/java/com/teradata/tempto/internal/ReflectionInjectorHelper.java
```diff
@@ -0,0 +1,76 @@
...
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Method;
+import java.lang.reflect.Parameter;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Predicate;
+
+import static java.util.Arrays.asList;
+
+public class ReflectionInjectorHelper
```
losipiuk: As we talked maybe better approach would be to create bytecode for class with field for each injectable method parameter and then run `injectMembers` on the class. This way you do not have to replicate Guice logic by hand.

PR #58, opened 2015-10-01
runner: introduce runTempto method which accepts command args

kokosing:
runner: introduce runTempto method which accepts command args

Thanks to that is super easy to write default main method for runner
class.

Test Plan: build, use in presto

Reviewers: losipiuk


losipiuk:
looks good

PR #64, opened 2015-10-07
examples: remove failing test

kokosing:
examples: remove failing test

Having failing test failing in automation may hide other failures.

Task: SWARM-1404

Test Plan: build

Reviewers: losipiuk


losipiuk:
do not remove this. 
It was there on purpose to be able to see how tf behaves in case of failing tests.
just exclude "failing" group in automation.

PR #65, opened 2015-10-14
tempto-core: add closeConnection and openConnections methods to JdbcQueryExecutor

arhimondr:
Useful for the cases when some test or fulfiller breaks the connection and we need to recover it.

Testing: unit tests


losipiuk:
I do not get the motivation. Let's talk offline

PR #78, opened 2015-11-23
Map binary values as binary not boolean.

kokosing:
Map binary values as binary not boolean.

Testing: unit tests

Reviewers: losipiuk, akshat

### DIFF
--- a/tempto-core/src/test/groovy/com/teradata/tempto/internal/query/QueryRowMapperTest.groovy
+++ b/tempto-core/src/test/groovy/com/teradata/tempto/internal/query/QueryRowMapperTest.groovy
@@ -1,0 +1,57 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


losipiuk:
looks good

tempto-core/src/test/groovy/com/teradata/tempto/internal/query/QueryRowMapperTest.groovy
```diff
@@ -0,0 +1,57 @@
...
+  {
+    setup:
+    QueryRowMapper rowMapper = new QueryRowMapper(ImmutableList.of(JDBCType.BINARY))
+
+    expect:
+    rowMapper.mapToRow(ImmutableList.of(value)).getValues()[0] == expected
+
+    where:
+    value  | expected
+    '0000' | bytes(0x00, 0x00)
```
akshatnair: Does this work if it is not even number of digits like 1a3, the user might enter this instead of 01a3?

akshatnair:
LGTM, will try it out with TD tests once this is pushed.

tempto-core/src/test/groovy/com/teradata/tempto/internal/query/QueryRowMapperTest.groovy
kokosing: number of digits has to even, I will add separate tests case for odd number of digits

PR #79, opened 2015-11-23
tempto-core: ensure that hive tables are created as external

arhimondr:
DROP table for not external tabled deletes all the table data.
As we want to keep table data between the product tests runs we have to ensure that all the hive tables are created as EXTERNAL.
Although Hive tables declared in TpchTableDefinitions are created as external, some user-declared table could be create with just plain "CREATE TABLE" statement instead of "CREATE EXTERNAL TABLE".
In such case after 2 tests runs in line such tables will be empty.

First iteration: UPLOAD DATA -> DROP TABLE IF EXISTS (nothing to drop at first iteration) -> CREATE TABLE
Second iteration: UPLOAD DATA (already there, skip) -> DROP IF EXIST(drops table and the table data) -> CREATE TABLE -> now we have table without any data.

To fix that we must ensure that all Hive tables are EXTERNAL.


kokosing:
lgtm

PR #80, opened 2015-11-26
Table definition (requirement) can specify in schema.

### LINKED ISSUE #76: Unable to specify database schema in tempto
Currently the only way to specify a schema to use is to append it to the jdbc url in the configuration file.   This means that you need to define a different database for each schema that you want to use for a given connection.  It would be useful if in the functions to create and access tables you could specify the schema to create tables in for when you don't want to use the default.

kokosing:
Table definition (requirement) can specify in schema.

Now it is possible to specify a schema for table definition in table
requirement. When it is specified given table will be created in that
schema. Note that to access that table you from other schema you would
need to prefix it with schema.

Schema can be also specified for table via convention ddl file.

See examples of how to use it.

Introduced two concepts into the codebase:
 - table handle - kind of a 'pointer' to table, it can be used in search
 of a table to tell where table is stored or in create to tell where
 this table should be stored
 - table name - contains exact place of where a real table instance is
 stored

Fixes #76.

Task: SWARM-1753
Testing: unit tests, running examples
Reviewers: losipiuk


losipiuk:
some comments - let's talk offline

kokosing:
Comments addressed.

losipiuk:
one comment

PR #86, opened 2016-01-05
Don't drop and recreate immutable JDBC tables

cawallin:
JDBC tables take a long time to reload, since they are loaded via prepared
statements, so don't drop them every time. A future extension would be to
drop and recreate the table if the underlying data has changed.

### DIFF
--- a/tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
+++ b/tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
@@ -15,6 +15,7 @@
 package com.teradata.tempto.internal.fulfillment.table.jdbc;
 
 import com.google.inject.Inject;
+import com.teradata.tempto.configuration.Configuration;
 import com.teradata.tempto.fulfillment.table.MutableTableRequirement.State;
 import com.teradata.tempto.fulfillment.table.TableDefinition;
 import com.teradata.tempto.fulfillment.table.TableHandle;
@@ -27,6 +28,7 @@ import com.teradata.tempto.internal.fulfillment.table.TableName;
 import com.teradata.tempto.internal.fulfillment.table.TableNameGenerator;
 import com.teradata.tempto.query.QueryExecutionException;
 import com.teradata.tempto.query.QueryExecutor;
+import com.teradata.tempto.query.QueryResult;
 import org.slf4j.Logger;
 
 import javax.inject.Named;
@@ -39,6 +41,7 @@ import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Optional;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
@@ -62,15 +65,18 @@ public class JdbcTableManager
 
     private final QueryExecutor queryExecutor;
     private final String databaseName;
+    private final Configuration configuration;
 
     @Inject
     public JdbcTableManager(
             QueryExecutor queryExecutor,
             TableNameGenerator tableNameGenerator,
-            @Named("databaseName") String databaseName)
+            @Named("databaseName") String databaseName,
+            Configuration configuration)
     {
         super(queryExecutor, tableNameGenerator);
         this.databaseName = databaseName;
+        this.configuration = configuration;
         this.queryExecutor = checkNotNull(queryExecutor, "queryExecutor is null");
     }
 
@@ -84,6 +90,8 @@ public class JdbcTableManager
             createTable(tableDefinition, tableName);
             JdbcTableDataSource dataSource = tableDefinition.getDataSource();
             insertData(tableName, dataSource);
+        } else {
+            LOGGER.info("Table {} already exists, skipping creation of immutable table", tableName.getNameInDatabase());
         }
         return new JdbcTableInstance(tableName, tableDefinition);
     }
@@ -100,18 +108,12 @@ public class JdbcTableManager
                     escapeNamePattern(tableName.getName(), escape),
                     new String[] {"TABLE"})
             ) {
-                while (resultSet.next()) {
-                    if (tableNameMatches(tableName, resultSet) && tableSchemaMatches(tableName, resultSet)) {
-                        LOGGER.info("Table {} already exists", tableName.getNameInDatabase());
-                        return true;
-                    }
-                }
+                return QueryResult.forResultSet(resultSet).getRowsCount() > 0;
             }
         }
         catch (SQLException e) {
             throw new QueryExecutionException(e);
         }
-        return false;
     }
 
     private static String escapeNamePattern(String name, String escape)
@@ -127,28 +129,14 @@ public class JdbcTableManager
         return name;
     }
 
-    private boolean tableNameMatches(TableName tableName, ResultSet resultSet)
-            throws SQLException
-    {
-        return resultSet.getString("TABLE_NAME").toLowerCase(ENGLISH).equals(tableName.getNameInDatabase().toLowerCase());
-    }
-
-    private boolean tableSchemaMatches(TableName tableName, ResultSet resultSet)
-            throws SQLException
-    {
-        if (tableName.getSchema().isPresent()) {
-            return resultSet.getString("TABLE_SCHEM").toLowerCase(ENGLISH).equals(tableName.getSchema().get().toLowerCase());
-        }
-        else {
-            return true;
-        }
-    }
-
     private void createTable(JdbcTableDefinition tableDefinition, TableName tableName)
     {
-        tableName.getSchema().ifPresent(schema ->
-                        queryExecutor.executeQuery("CREATE SCHEMA IF NOT EXISTS " + schema)
-        );
+        boolean skipCreateSchema = configuration.getBoolean("databases." + databaseName + ".skip_create_schema").orElse(false);
+        if (!skipCreateSchema) {
+            tableName.getSchema().ifPresent(schema ->
+                            queryExecutor.executeQuery("CREATE SCHEMA IF NOT EXISTS " + schema)
+            );
+        }
 
         queryExecutor.executeQuery(tableDefinition.getCreateTableDDL(tableName.getNameInDatabase()));
     }
--- a/tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
+++ b/tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
@@ -25,17 +25,22 @@ import com.teradata.tempto.fulfillment.table.jdbc.JdbcTableDefinition;
 import com.teradata.tempto.internal.fulfillment.table.AbstractTableManager;
 import com.teradata.tempto.internal.fulfillment.table.TableName;
 import com.teradata.tempto.internal.fulfillment.table.TableNameGenerator;
+import com.teradata.tempto.query.QueryExecutionException;
 import com.teradata.tempto.query.QueryExecutor;
 import org.slf4j.Logger;
 
 import javax.inject.Named;
 
+import java.sql.Connection;
+import java.sql.DatabaseMetaData;
 import java.sql.PreparedStatement;
+import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.collect.Lists.newArrayList;
 import static com.teradata.tempto.fulfillment.table.MutableTableRequirement.State.CREATED;
@@ -43,6 +48,7 @@ import static com.teradata.tempto.fulfillment.table.MutableTableRequirement.Stat
 import static com.teradata.tempto.fulfillment.table.MutableTableRequirement.State.PREPARED;
 import static java.lang.String.format;
 import static java.util.Collections.unmodifiableList;
+import static java.util.Locale.ENGLISH;
 import static java.util.stream.Collectors.joining;
 import static java.util.stream.IntStream.range;
 import static org.slf4j.LoggerFactory.getLogger;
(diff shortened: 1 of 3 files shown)


tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
```diff
@@ -73,18 +79,77 @@ public TableInstance createImmutable(JdbcTableDefinition tableDefinition, TableH
...
+            createTable(tableDefinition, tableName);
+            JdbcTableDataSource dataSource = tableDefinition.getDataSource();
+            insertData(tableName, dataSource);
+        }
         return new JdbcTableInstance(tableName, tableDefinition);
     }
 
+    private boolean tableExists(TableName tableName)
+    {
+        Connection connection = queryExecutor.getConnection();
```
cawallin: Theoretically we could move tableExists() to queryExecutor, but it seemed a bit too specific to me to be in that interface.

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
```diff
@@ -73,18 +79,77 @@ public TableInstance createImmutable(JdbcTableDefinition tableDefinition, TableH
     {
         TableName tableName = createImmutableTableName(tableHandle);
         LOGGER.debug("creating immutable table {}", tableName);
-        dropTableIgnoreError(tableName); // ultimately we do not want to drop table here - if data did not change
-        createTable(tableDefinition, tableName);
-        JdbcTableDataSource dataSource = tableDefinition.getDataSource();
-        insertData(tableName, dataSource);
+        // TODO: drop and recreate table if the underlying data has changed
+        if (!tableExists(tableName)) {
+            createTable(tableDefinition, tableName);
```
maciejgrzybek: Can you maybe extract the following three lines to a separate method?
Maybe `private void createTableWithData(tableDefinition, tableName)`?

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
maciejgrzybek: try-with-resources for `connection`?

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
```diff
@@ -73,18 +79,77 @@ public TableInstance createImmutable(JdbcTableDefinition tableDefinition, TableH
...
+        name = name.replace(escape, escape + escape);
+        name = name.replace("_", escape + "_");
+        name = name.replace("%", escape + "%");
+        return name;
+    }
+
+    private boolean tableNameMatches(TableName tableName, ResultSet resultSet)
+            throws SQLException
+    {
+        return resultSet.getString("TABLE_NAME").toLowerCase(ENGLISH).equals(tableName.getNameInDatabase());
```
maciejgrzybek: `toLowerCase(tableName.getNameInDatabase())` for consistency and robustness?

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
```diff
@@ -73,18 +79,77 @@ public TableInstance createImmutable(JdbcTableDefinition tableDefinition, TableH
...
+    }
+
+    private boolean tableSchemaMatches(TableName tableName, ResultSet resultSet)
+            throws SQLException
+    {
+        if (tableName.getSchema().isPresent()) {
+            return resultSet.getString("TABLE_SCHEM").toLowerCase(ENGLISH).equals(tableName.getSchema().get());
```
maciejgrzybek: ditto `toLowerCase`

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
maciejgrzybek: :+1: `tableExists()` should not be in `QueryExecutor` to me.

cawallin: we don't want to close the connection afterwards, since the single connection is used for all communication with Teradata

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
```diff
@@ -73,18 +79,77 @@ public TableInstance createImmutable(JdbcTableDefinition tableDefinition, TableH
     {
         TableName tableName = createImmutableTableName(tableHandle);
         LOGGER.debug("creating immutable table {}", tableName);
-        dropTableIgnoreError(tableName); // ultimately we do not want to drop table here - if data did not change
-        createTable(tableDefinition, tableName);
-        JdbcTableDataSource dataSource = tableDefinition.getDataSource();
-        insertData(tableName, dataSource);
+        // TODO: drop and recreate table if the underlying data has changed
```
akshatnair: I think this can be based off of the version number?

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
```diff
@@ -73,18 +79,77 @@ public TableInstance createImmutable(JdbcTableDefinition tableDefinition, TableH
...
+                }
+            }
+        }
+        catch (SQLException e) {
+            throw new QueryExecutionException(e);
+        }
+        return false;
+    }
+
+    protected static String escapeNamePattern(String name, String escape)
```
akshatnair: protected needed?

akshatnair:
lgtm mod comment

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
losipiuk: I would rather use query on "information_schema" which is ANSI standard.

```sql
SELECT EXISTS (
   SELECT 1
   FROM   information_schema.tables 
   WHERE  table_schema = 'schema_name'
   AND    table_name = 'table_name'
);
```

cawallin: Teradata doesn't have an information_schema table... it uses dbc instead. I guess we could make that configurable or something.

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
cawallin: nope, removed

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
cawallin: done

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
cawallin: done

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
cawallin: But for the TPCH tables, there is no version number? I'm going to leave this be for now, I think it means adding some additional infrastructure.

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
losipiuk: no - if we need to use metadata for some cases - let's leave it this way. I was hoping for code simplification.

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
akshatnair: Ah okay, we should file an issue for its support then

cawallin:
Added schema so that we can know when to create the table in Teradata (which previously had not been able to specify the schema).  However, mutable tables are still not working with it.

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/TableName.java
```diff
@@ -26,13 +26,20 @@
...
     private final String name;
     private final String nameInDatabase;
+    private final boolean shouldCreateSchema;
 
     public TableName(String database, Optional<String> schema, String name, String nameInDatabase)
     {
+        this(database, schema, name, nameInDatabase, true);
+    }
+
+    public TableName(String database, Optional<String> schema, String name, String nameInDatabase, boolean shouldCreateSchema)
```
maciejgrzybek: Can't we get rid of this bool argument? Maybe through derivation? When you see series of true/false it's not readable what they stand for without looking into details of the class.

cawallin: Hm, namely having two classes? Adding the boolean arg is a bit hacky in general

cawallin:
Updated so that mutable tables work

tempto-core/src/main/java/com/teradata/tempto/fulfillment/table/TableHandle.java
```diff
@@ -84,6 +91,10 @@ public TableHandle withNoSchema()
        return new TableHandle(database, Optional.empty(), name, true);
     }
 
+    public TableHandle shouldNotCreateSchema() {
```
losipiuk: I would rather replace this with shouldCreateSchema(boolean)

tempto-core/src/main/java/com/teradata/tempto/fulfillment/table/jdbc/JdbcTableDefinition.java
```diff
@@ -105,6 +108,12 @@ public JdbcTableDefinitionBuilder withSchema(String schema)
             return this;
         }
 
+        public JdbcTableDefinitionBuilder shouldNotCreateSchema()
```
losipiuk: ditto

tempto-core/src/main/java/com/teradata/tempto/fulfillment/table/jdbc/JdbcTableDefinition.java
```diff
@@ -85,6 +85,9 @@ public static JdbcTableDefinitionBuilder like(JdbcTableDefinition tableDefinitio
         if (tableDefinition.getSchema().isPresent()) {
             builder.withSchema(tableDefinition.getSchema().get());
         }
+        if (tableDefinition.getTableHandle().getShouldCreateSchema()) {
```
losipiuk: missing `!` ?

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/TableName.java
losipiuk: maybe this should be set on the database level - not on the tableDefinition level.
Is it that we want to create schema for some TD tables and not create for others?

cawallin: ahh, that would work. We never want to create schema for TD tables. I'll switch it to that.

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
```diff
@@ -73,18 +79,79 @@ public TableInstance createImmutable(JdbcTableDefinition tableDefinition, TableH
...
+            String escape = metadata.getSearchStringEscape();
+            try (ResultSet resultSet = metadata.getTables(
+                    connection.getCatalog(),
+                    escapeNamePattern(tableName.getSchema().orElse(null), escape),
+                    escapeNamePattern(tableName.getName(), escape),
+                    new String[] {"TABLE"})
+            ) {
+                while (resultSet.next()) {
+                    if (tableNameMatches(tableName, resultSet) && tableSchemaMatches(tableName, resultSet)) {
+                        LOGGER.info("Table {} already exists", tableName.getNameInDatabase());
```
losipiuk: move this `LOGGER.info` statement to `createImmutable` and add `" skipping creating immutable table"` to log message.

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
```diff
@@ -73,18 +79,79 @@ public TableInstance createImmutable(JdbcTableDefinition tableDefinition, TableH
...
+        return name;
+    }
+
+    private boolean tableNameMatches(TableName tableName, ResultSet resultSet)
+            throws SQLException
+    {
+        return resultSet.getString("TABLE_NAME").toLowerCase(ENGLISH).equals(tableName.getName().toLowerCase());
+    }
+
+    private boolean tableSchemaMatches(TableName tableName, ResultSet resultSet)
```
losipiuk: why do you specify collation for one string and not for the other?

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
```diff
@@ -73,18 +79,79 @@ public TableInstance createImmutable(JdbcTableDefinition tableDefinition, TableH
...
+            throws SQLException
+    {
+        if (tableName.getSchema().isPresent()) {
+            return resultSet.getString("TABLE_SCHEM").toLowerCase(ENGLISH).equals(tableName.getSchema().get().toLowerCase());
```
losipiuk: ditto

losipiuk:
@kokosing please take a look as you played with this codebase most recently.

tempto-core/src/main/java/com/teradata/tempto/fulfillment/table/TableHandle.java
```diff
@@ -50,33 +50,40 @@ public static TableHandle parse(String value)
     private final Optional<String> schema;
     private final String name;
     private final boolean requireNoSchema;
+    private final boolean shouldCreateSchema;
```
kokosing: I don't think it is a best place for this field. TableHandle tells you what table should be used, so tempto is able to find it among different databases and schemas. It is not only used for creation but also for table selection. In case of selection this field has no sense.

tempto-core/src/main/java/com/teradata/tempto/fulfillment/table/TableHandle.java
```diff
@@ -99,6 +110,10 @@ public String getName()
         return schema;
     }
 
+    public boolean getShouldCreateSchema() {
```
kokosing: accessors for booleans should start from `is`. In this case I would get rid of `get` or renamed it to `isSchemaCreationNeeded()`

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
kokosing: what about using jdbc metadata? see https://docs.oracle.com/javase/7/docs/api/java/sql/DatabaseMetaData.html#getTables(java.lang.String,%20java.lang.String,%20java.lang.String,%20java.lang.String[])

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
```diff
@@ -73,18 +79,79 @@ public TableInstance createImmutable(JdbcTableDefinition tableDefinition, TableH
...
+        try {
+            DatabaseMetaData metadata = connection.getMetaData();
+            String escape = metadata.getSearchStringEscape();
+            try (ResultSet resultSet = metadata.getTables(
+                    connection.getCatalog(),
+                    escapeNamePattern(tableName.getSchema().orElse(null), escape),
+                    escapeNamePattern(tableName.getName(), escape),
+                    new String[] {"TABLE"})
+            ) {
+                while (resultSet.next()) {
```
kokosing: maybe use QueryResult here

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
```diff
@@ -73,18 +79,79 @@ public TableInstance createImmutable(JdbcTableDefinition tableDefinition, TableH
...
+        name = name.replace(escape, escape + escape);
+        name = name.replace("_", escape + "_");
+        name = name.replace("%", escape + "%");
+        return name;
+    }
+
+    private boolean tableNameMatches(TableName tableName, ResultSet resultSet)
+            throws SQLException
+    {
+        return resultSet.getString("TABLE_NAME").toLowerCase(ENGLISH).equals(tableName.getName().toLowerCase());
```
kokosing: maybe use equalsIgnoreCase?

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/TableName.java
```diff
@@ -26,13 +26,20 @@
     private final Optional<String> schema;
     private final String name;
     private final String nameInDatabase;
+    private final boolean shouldCreateSchema;
```
kokosing: remove it from here, TableName is about existing tables in database, it has nothing to do with table creation

tempto-core/src/main/java/com/teradata/tempto/fulfillment/table/TableHandle.java
kokosing: I think the better place is TableDefinition

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
cawallin: ah, yes of course :)

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
cawallin: helpful, thanks!

cawallin:
Much improved version, thanks for the review @losipiuk and @kokosing. I switched it to be a property you set in the database config.

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
```diff
@@ -73,18 +85,59 @@ public TableInstance createImmutable(JdbcTableDefinition tableDefinition, TableH
...
+        name = name.replace("%", escape + "%");
+        return name;
+    }
+
     private void createTable(JdbcTableDefinition tableDefinition, TableName tableName)
     {
-        tableName.getSchema().ifPresent(schema ->
-                        queryExecutor.executeQuery("CREATE SCHEMA IF NOT EXISTS " + schema)
-        );
+        boolean skipCreateSchema = configuration.getBoolean("databases." + databaseName + ".skip_create_schema").orElse(false);
```
kokosing: add information about that to README

kokosing:
lgtm

PR #87, opened 2016-01-11
Drop immutable table if no schema is specified

cawallin:
If no schema is specified, we have to drop and recreate immutable tables because
we don't know if the table already exists or if it's the same table name
in a different schema.

### DIFF
--- a/tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
+++ b/tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
@@ -86,14 +86,25 @@ public class JdbcTableManager
         TableName tableName = createImmutableTableName(tableHandle);
         LOGGER.debug("creating immutable table {}", tableName);
         // TODO: drop and recreate table if the underlying data has changed
-        if (!tableExists(tableName)) {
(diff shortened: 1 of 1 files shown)


cawallin:
With this, the tempto examples all pass. @kokosing  @losipiuk

kokosing:
code is ok, but I don't understand the motivation (problem)

from commit message `the same table name in a different schema`, the semantic of table name is to have exact information (no optionals) about table (db, schema, name) so in my understanding it is impossible to have the same table name in two different schemas as, because they are in different schemas so they are different

cawallin:
The problem is that if you have a nation table in the default schema (with schema specified as optional) and nation table in the tpch schema, the nation table in the default schema will not be created because when you getTables() from the metastore there is already a nation table, so you don't need to recreate the table. Alternatively, we can force schema to be specified all the time, but that seemed like a more wide-ranging change.

kokosing:
Can you check what is the default schema, then you could use it for tables returned from getTables(). If returned table is in schema which is your default you can remove it for table with no schema specified. How does it sound? 

PS. write to me on slack, when you write your response so I will be able answer within the same day

cawallin:
There doesn't seem to be a generic way to get the default schema:  http://stackoverflow.com/questions/20238520/how-to-get-the-default-schema-of-a-sql-connection.  So as you agreed on slack, I'm going to push this as is.

PR #88, opened 2016-01-13
Fixes negative double comparison by using positive tolerance value

kkc00:
Previously the tolerance for DoubleMath.fuzzyCompare() used input derived value which resulted in negative tolerance if negative double/float values were compared.  DoubleMath.fuzzyCompare() throws exception if the precision value is negative, therefore comparing negative double/float fails. 

This patch uses a positive tolerance value. The tolerance value is taken from IEEE 754 roundoff error.

### DIFF
--- a/tempto-core/src/main/java/com/teradata/tempto/internal/query/QueryResultValueComparator.java
+++ b/tempto-core/src/main/java/com/teradata/tempto/internal/query/QueryResultValueComparator.java
@@ -39,7 +39,7 @@ public class QueryResultValueComparator
         implements Comparator<Object>
 {
 
-    private static final double DOUBLE_FUZZY_MATCH_THRESHOLD = 0.00000000001;
+    private static final double MACHINE_EPSILON_DOUBLE_64B = Math.pow(2,-53);
     private final JDBCType type;
 
     private QueryResultValueComparator(JDBCType type)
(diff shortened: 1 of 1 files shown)


tempto-core/src/main/java/com/teradata/tempto/internal/query/QueryResultValueComparator.java
```diff
@@ -188,7 +188,7 @@ private int doubleEqual(Object actual, Object expected)
         if (isFloatingPointValue(actual) && isFloatingPointValue(expected)) {
             Double actualDouble = Double.valueOf(actual.toString());
             Double expectedDouble = Double.valueOf(expected.toString());
-            Double threshold = expectedDouble * DOUBLE_FUZZY_MATCH_THRESHOLD;
+            Double threshold = DOUBLE_FUZZY_MATCH_THRESHOLD;
```
cawallin: Inline this variable

cawallin:
looks good % comment (but wait for @kokosing or @losipiuk to review it in case there was some other reasoning behind this comparator)

tempto-core/src/main/java/com/teradata/tempto/internal/query/QueryResultValueComparator.java
```diff
@@ -39,7 +39,7 @@
         implements Comparator<Object>
 {
 
-    private static final double DOUBLE_FUZZY_MATCH_THRESHOLD = 0.00000000001;
+    private static final double DOUBLE_FUZZY_MATCH_THRESHOLD = Math.pow(2,-53);
```
maciejgrzybek: Rename this constant to something more descriptive, e.g. `MACHINE_EPSILON_DOUBLE_64B`.

tempto-core/src/main/java/com/teradata/tempto/internal/query/QueryResultValueComparator.java
kkc00: done in the updated commit.

tempto-core/src/main/java/com/teradata/tempto/internal/query/QueryResultValueComparator.java
kkc00: done in the updated commit.

kokosing:
looks good to me, please wait for @losipiuk  to review

losipiuk:
looks good

rschlussel:
merged.

PR #89, opened 2016-01-13
Timestamp fix

kkc00:
creating a new timestamp class that prints timestamp in UTC instead of local

Edit: providing a description, had communicated with others before. 

Tested for "timestamp" field with no timezone information. 
Following example scenario happens currently: 
1. input in data file (.data file) is "2014-02-01 23:59:21"
2. tempto creates a Timestamp object using DateTimeUtils.parseTimestampInUTC(), the processing is correct. 
3. tempto inserts Timestamp object to teradata, using Timestamp.toString() value
4. But, Timestamp.toString() prints "2014-02-01 18:59:21.0" which is local equivalent in EST
5. Database stores "2014-02-01 18:59:21.0"
6. select query via tempto when using onPresto() returns "2014-02-01 13:59:21.0", which is due to Presto converting the timestamp from UTC to local (EST). 
7. So, final value returned by tempto is "2014-02-01 13:59:21.0". This value is 2 timezones  (2 x 5hours offset for EST) away from the original value of  "2014-02-01 23:59:21"

So the consensus was, if in step 5, if the database contains original timestamp i.e.  "2014-02-01 23:59:21", then we can do correct output comparison. 

More investigation showed that the Timestamp object in DateTimeUtils.parseTimestampInUTC() processes the UTC timezone information correctly, but when printing by its toString() function, it prints the local timestamp.

Therefore, the patch contains a Timestamp class definition that overrides Timestamp.toString() function to print the correct timestamp to be inserted to database. 

Something similar is discussed here also: 
http://stackoverflow.com/questions/18604989/utc-timestamp-joda-time

### DIFF
--- a/tempto-core/src/main/java/com/teradata/tempto/internal/initialization/TestInitializationListener.java
+++ b/tempto-core/src/main/java/com/teradata/tempto/internal/initialization/TestInitializationListener.java
@@ -53,6 +53,7 @@ import java.lang.reflect.Method;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.TimeZone;
 
 import static com.beust.jcommander.internal.Lists.newArrayList;
 import static com.google.common.base.Preconditions.checkState;
@@ -172,6 +173,7 @@ public class TestInitializationListener
     @Override
     public void onStart(ITestContext context)
     {
+        TimeZone.setDefault(TimeZone.getTimeZone("UTC"));
         Module suiteModule = combine(combine(getSuiteModules()), bind(suiteLevelFulfillers), bind(testMethodLevelFulfillers));
         GuiceTestContext initSuiteTestContext = new GuiceTestContext(suiteModule);
         TestContextStack<GuiceTestContext> suiteTextContextStack = new TestContextStack<>();


akshatnair:
It will be useful if the commit message also gives scenario where the Timestamp class was not providing the benefit we are now receiving by adding this new class. And commit header is not very informativ, like instead of Timestamp fix (you are not fixing anything in timestamp) it should be Added UTCTimeStamp class

kkc00:
@akshatnair added description above.

cawallin:
Actually, the first line of the commit message should be something that explains the problem and fix -- Add UtcTimeStamp class so that timestamps are handled in UTC time zone

akshatnair:
@kkc00 thanks for the updated info

kokosing:
looks good

losipiuk:
In point **(3)** you say that when `Timestamp` value is inserted into Teradata `toString()` method is called.
Where does that happen? In tempto or is it part of internal implementation of teradata's jdbc driver.

It seems to me that latter is the case.

From what I see we just use the `setObject(col, Timestamp)` method on `PreparedStatement` to insert data into table in `JdbcTableManager`.

If what I wrote above is true then I do not believe this patch is proper approach to fix current problems. 
Instead we should either:
 * treat `TIMESTAMP WITHOUT TIMEZONE` columns differently in tempto when inserting data and not use `PreparedStatement.setObject` method for them.
 * configure Teradata's JDBC session in such way that it interpret `Timestamp` object passed to `setObject` method in UTC timezone.

kkc00:
@losipiuk I have made changes closer to your second suggestion, but it is about setting the default timezone to UTC. The default timezone value is used by the Timestamp object to format its output in Timestamp.toString(). This aligns with tempto's behavior that every timestamp without timezone is treated as UTC timestamp and passed through DateTimeUtils.parseTimestampInUTC() via QueryRowMapper.convertValue(). This change fixes the issue outlined in the original post.

kkc00:
@losipiuk Moved the setting of default timezone to TestInitializationListener as discussed. Let me know if this is alright.

losipiuk:
Seems fine - check if presto tests pass after the change. 
Per Christina comment. We should still be able to test database for any timezone. Kamal's change how client (tempto) VM interprets Timestamp literals when printing them out etc. And how database interprets literal columns is matter of session configuration. Right?

cawallin:
It just means that the tempto client will only be in UTC right? I don't think it's a huge problem, but we should at least be aware of it.

kkc00:
@losipiuk  Yes. That is the effect of having a default timezone. Timestamp object (and Date object too) by itself does not have any timezone. It only stores the number of milliseconds since  January 1, 1970, 00:00:00 GMT and Timestamp allows more granularity (with respect to Date object). So, while printing the Timestamp in date format (MM-DD-YY hh-mm-ss or others), it needs assistance in what to print, as the string to be printed in different timezone will differ. Tempto already treats timestamp without timezone as being in UTC, which is the correct behavior, so this change will make sure that the string passed to the database is same as the one present in the .data file or other row based input. How the database treats the string is dependent upon the database. 

As for supporting other timezones, that is related to parsing the timestamp with timezone strings that are present in .data file or manually specified by row objects. Though that is not supported currently in .data file, support for it can be added by introducing more patterns in DatetimeUtils.java. So, this change does not affect whether tempto supports other timezone timestamps or not. 

I will check the presto tests.

kkc00:
@losipiuk  Presto product tests were run using this fix. It has no effect on existing behavior of the tests except for testing the output timestamp in UTC format instead of the default of Asia/Katmandu that happens currently. 

The timestamp string sent to hive is accurate, with or without this change (might be because of different way in which data is loaded to hive vs. teradata database). But, it looks like the way hive interprets timestamp string (meaning: what the value of the timestamp milleseconds will be) is based on its own local timezone. This is different from how teradata handles timestamp strings. Both of these behaviors are equally ok, as it is expected that databases can behave differently on how they interpret timestamp strings without timezone (mentioned in previous post as well). 

Therefore, as a conclusion on the purpose of this patch, it makes tempto consistent on how a given timestamp string that does not have any timezone information is interpreted and how a given  timestamp millisecond value is printed in string format. There might be other ways to influence hive behavior via tempto (other than the configs). I might look at it later, but, for now it looks like out of scope for this fix. 

I do not have commit access on tempto, therefore, let me know if you would be ok with applying this.

losipiuk:
> The timestamp string sent to hive is accurate, with or without this change (might be because of different way in which data is loaded to hive vs. teradata database). 

Yeah - for hive we just copy the file to HDFS. So no interpretation of data in file is performed on presto side.

Your patch seems fine to me. Push it - if we later see that it breaks sth - we can always change it again.

cawallin:
Merged

PR #91, opened 2016-01-20
Fix empty batch insert issue

anusudarsan:
Insert into Teradata would fail when calling ps.executeBatch() on an
emtpy batch.
https://forums.teradata.com/forum/connectivity/terajdbc-13-10-00-32-error-1382-sqlstate-hy000-the-batch-is-empty. So fix this in tempto.


cawallin:
looks good!

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/JdbcTableManager.java
```diff
@@ -1,4 +1,5 @@
 /*
+/*
```
maciejgrzybek: Too many `/*`s, isn't it?

PR #93, opened 2016-02-10
runner: add possibility for user to specify own parameters

kokosing:
runner: add possibility for user to specify own parameters


arhimondr:
lgtm

PR #95, opened 2016-02-15
Handle Xattrs were attribute is not set.

kokosing:
Handle Xattrs were attribute is not set.


cawallin:
Does this handle the following issue? https://jira.td.teradata.com/jira/browse/SWARM-1838 (underlying https://issues.apache.org/jira/browse/HDFS-7997)

kokosing:
@cawallin yes it does.

pnowojski:
lgtm

PR #96, opened 2016-02-16
Add possibility to define tolerance for float value comparisons

kokosing:
Add possibility to define tolerance for float value comparisons

Reviewers: losipiuk

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,7 +223,14 @@ This section is used to configure various properties used during test execution.
 tests:
   hdfs:
     path: /tempto  # where to store test data on HDFS
+  assert:
+    float_tolerance: 0.0001
 ```
+
+| property | description |
+|----------|-------------|
+| tests.hdfs.path | defines where data for tables will be stored in hdfs |
+| tests.assert.float_tolerance | defines tolerance for floating point values comparision |
 
 ## Java based tests
 
--- a/tempto-core/src/main/java/com/teradata/tempto/internal/query/QueryResultValueComparator.java
+++ b/tempto-core/src/main/java/com/teradata/tempto/internal/query/QueryResultValueComparator.java
@@ -17,6 +17,7 @@ package com.teradata.tempto.internal.query;
 import com.google.common.base.Throwables;
 import com.google.common.math.DoubleMath;
 import com.google.common.primitives.UnsignedBytes;
+import com.teradata.tempto.configuration.Configuration;
 
 import java.math.BigDecimal;
 import java.sql.Array;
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -223,8 +223,17 @@ This section is used to configure various properties used during test execution.
 tests:
   hdfs:
     path: /tempto  # where to store test data on HDFS
+  assert:
+    diff:
```
losipiuk: maybe replace `diff: ... float: ... tolearance:` with just one level `float_tolerance:`

tempto-core/src/main/java/com/teradata/tempto/internal/query/QueryResultValueComparator.java
```diff
@@ -196,12 +203,21 @@ private static int floatingEqualWithTolerance(Object actual, Object expected, do
 
     private int singleFloatingEqual(Object actual, Object expected)
     {
-        return floatingEqualWithTolerance(actual, expected, MACHINE_EPSILON_FLOATING_32B);
+        return floatingEqualWithTolerance(actual, expected, getFloatTolerance(MACHINE_EPSILON_FLOATING_32B));
```
losipiuk: remove MACHINE_EPSILON_FLOATING_32B - as those seem to not make sense as they do not depend on actual/expected value.

losipiuk: make tolerance passed to `floatingEqualWithTolerance` be a maximum allowed fraction of expected value. Not absolute allowed variance.

losipiuk:
looks good.

tempto-core/src/main/java/com/teradata/tempto/internal/query/QueryResultValueComparator.java
```diff
@@ -186,22 +191,19 @@ private int longEqual(Object actual, Object expected)
...
-    private int singleFloatingEqual(Object actual, Object expected)
-    {
-        return floatingEqualWithTolerance(actual, expected, MACHINE_EPSILON_FLOATING_32B);
-    }
 
-    private int doubleEqual(Object actual, Object expected)
-    {
-        return floatingEqualWithTolerance(actual, expected, MACHINE_EPSILON_FLOATING_64B);
+        double expectedDouble = getDoubleValue(expected);
+        double tolerance = 0;
```
maciejgrzybek: I believe machine epsilon is a better default since tolerance = 0 does not have sense in double comparison by definition.

PR #99, opened 2016-02-29
Replace configuration variables with System.properties

kokosing:
Replace configuration variables with System.properties


losipiuk:
looks good

PR #109, opened 2016-03-31
Update help message

### LINKED ISSUE #103: --config-local should run with path instead of URI
Right now --help says that this flag takes URI as parameter. In fact the URI is accepted on Mac OS X and not on GNU/Linux. On the other hand only GNU/Linux accepts path as this parameter.
It would be nice to have that cleaned to avoid two different commands for two platforms and have a valid --help.

kokosing:
Update help message

Fixes #103


kokosing:
@fiedukow

maciejgrzybek:
Looks good. Does it solve #103? Have you checked on Mac with path (as help states) provided (instead of URI)?

fiedukow:
LGTM, looks like #103 is solved already.

PR #113, opened 2016-04-05
Add docker based cluster usage for running tempto-examples tests

kokosing:
Add docker based cluster usage for running tempto-examples tests


.gitignore
```diff
@@ -18,9 +18,6 @@ out/
 # tests logging output
 tempto_logs
 
-# tempto local configuration
```
sopel39: isn't it needed?

tempto-examples/bin/run_on_docker.sh
```diff
@@ -0,0 +1,88 @@
...
+  docker exec $(hadoop_master_container) hive -e 'show tables'
+}
+
+function docker_compose_network() {
+  docker network ls | grep docker | cut  -f 1 -d ' '
+}
+
+function check_presto() {
+  docker run \
+    --rm -it \
```
sopel39: interactive mode will fail on non-tty terminal. Ask @arhimondr for details

tempto-examples/bin/run_on_docker.sh
```diff
@@ -0,0 +1,88 @@
...
+    --net $(docker_compose_network) \
+    teradatalabs/presto-cli \
+    java -jar /presto-cli.jar --server presto-master:8080 \
+    --execute "SHOW CATALOGS"  | grep -i hive
+}
+
+function run_product_tests() {
+  docker run \
+    --rm -it \
```
sopel39: ditto

tempto-examples/docker/presto-cli/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM teradatalabs/centos6-java8-oracle
```
sopel39: where is this dockerfile used?

tempto-examples/bin/run_on_docker.sh
kokosing: Why should I care non-tty terminals? It designed to be run on travis or by developer locally.

tempto-examples/bin/run_on_docker.sh
```diff
@@ -0,0 +1,88 @@
...
+
+function docker_compose_network() {
+  docker network ls | grep docker | cut  -f 1 -d ' '
+}
+
+function check_presto() {
+  docker run \
+    --rm -it \
+    --net $(docker_compose_network) \
+    teradatalabs/presto-cli \
```
sopel39: can't you just launch command on running `presto-master` container? `presto-cli` image wouldn't be needed then

tempto-examples/docker/docker-compose.yml
```diff
@@ -0,0 +1,48 @@
...
+  presto-master:
+    hostname: presto-master
+    build: 'presto-server'
+    ports: 
+      - '8080:8080'
+    depends_on:
+      - 'hadoop-master'
+      - 'psql1'
+      - 'psql2'
+      - 'ssh'
```
sopel39: why `presto-master` depends on `ssh`?

tempto-examples/bin/run_on_docker.sh
sopel39: This is similar to what is in Presto so it would be good to make it work the same. I am not sure you can expect it to run only on Travis and by developer. It works on Travis (I am not sure if by design), but would fail on Jenkins.

tempto-examples/bin/run_on_docker.sh
kokosing: I don't want to mix responsibilities. Also presto-master does not have presto-jdbc jar or any client jars.

.gitignore
losipiuk: A comment?

tempto-examples/bin/run_on_docker.sh
```diff
@@ -0,0 +1,88 @@
...
+      break
+    fi
+    sleep 5
+  done
+
+  return ${EXIT_CODE}
+}
+
+function hadoop_master_container(){
+  docker ps --format '{{.Names}}' | grep hadoop-master
```
losipiuk: That will be better test:
```
docker ps -f name=hadoop-master -q |grep .
```

tempto-examples/bin/run_on_docker.sh
```diff
@@ -0,0 +1,88 @@
...
+}
+
+function check_hive() {
+  docker exec $(hadoop_master_container) hive -e 'show tables'
+}
+
+function docker_compose_network() {
+  docker network ls | grep docker | cut  -f 1 -d ' '
```
losipiuk: I do not have such network on mac.
Does it appear only if I run some docker compose?

Also I do not like this `grep`. It is very error prone. 
Can't we use `docker network inspect ...` instead?

.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+  - oraclejdk8
+
+sudo: required
+dist: trusty
+
+services:
+  - docker
+
+before_install:
+  # This section can be removed once travis get equipped with docker 1.10 and docker-compose 1.6
```
losipiuk: Is there an older version of docker on travis?
Or are we installing our own.
If the latter why cannot we install newer version?

tempto-examples/bin/run_on_docker.sh
```diff
@@ -0,0 +1,88 @@
...
+    --net $(docker_compose_network) \
+    teradatalabs/presto-cli \
+    java -jar /presto-cli.jar --server presto-master:8080 \
+    --execute "SHOW CATALOGS"  | grep -i hive
+}
+
+function run_product_tests() {
+  docker run \
+    --rm -it \
+    --net $(docker_compose_network) \
```
sopel39: why you don't just link connectors together by name instead of using `grep`?

build.gradle
```diff
@@ -43,7 +43,7 @@ ext.versions = [
         commons_lang3: '3.3.2',
         commons_dbcp2: '2.1',
         hive         : '0.13.1-3',
-        presto       : '0.97',
+        presto       : '0.143',
```
losipiuk: Unrelated change?

tempto-core/src/main/java/com/teradata/tempto/internal/listeners/ProgressLoggingListener.java
```diff
@@ -87,8 +86,15 @@ public void onTestSkipped(ITestResult testCase)
 
     private void logTestEnd(String outcome)
     {
-        long executionTime = System.currentTimeMillis() - testStartTime;
-        LOGGER.info("{}     /    took {}", outcome, DurationFormatUtils.formatDuration(executionTime, "m' minutes and 's' seconds'"));
+        if (LOGGER.isInfoEnabled()) {
+            long executionTime = System.currentTimeMillis() - testStartTime;
+            if (executionTime > 1000) {
```
losipiuk: why so?

tempto-examples/bin/run_on_docker.sh
```diff
@@ -0,0 +1,88 @@
...
+# try to stop already running containers
+docker-compose -f "${DOCKER_COMPOSE_LOCATION}" stop || true
+
+docker-compose -f "${DOCKER_COMPOSE_LOCATION}" pull
+docker-compose -f "${DOCKER_COMPOSE_LOCATION}" build
+
+docker-compose -f "${DOCKER_COMPOSE_LOCATION}" up -d 
+docker-compose -f "${DOCKER_COMPOSE_LOCATION}" logs --no-color &
+
+docker build -t teradatalabs/presto-cli "${PRODUCT_TESTS_ROOT}"/docker/presto-cli
```
losipiuk: don't we need to some cleanup after build here

tempto-examples/docker/presto-server/etc/catalog/jmx.properties
```diff
@@ -0,0 +1 @@
+connector.name=jmx
```
losipiuk: missing new line

losipiuk:
looks like an improvement :)

tempto-examples/bin/run_on_docker.sh
arhimondr: `docker-compose ps -q hadoop-master`

tempto-examples/bin/run_on_docker.sh
arhimondr: Please check this out https://github.com/Teradata/presto/pull/180

tempto-examples/bin/run_on_docker.sh
arhimondr: You can create yet another service in docker-compose, and run what you want with `docker-compose-run service-name`. So you don't need to determine a network and specify it for docker all around.

tempto-examples/bin/run_on_docker.sh
arhimondr: Shouldn't we add this image to the docker-compose file?

tempto-examples/bin/run_on_docker.sh
```diff
@@ -0,0 +1,88 @@
...
+retry check_hive
+retry check_presto
+
+# run product tests
+set +e
+run_product_tests "$*"
```
arhimondr: This will have problem with handling `INT` or `TERM` signals. You should make a trap.

build.gradle
kokosing: without this change I would need to use presto-server 0.97. I can extract this.

tempto-core/src/main/java/com/teradata/tempto/internal/listeners/ProgressLoggingListener.java
kokosing: It is meaningless to display to most of the tests that they last 0 minutes and 0 second.

tempto-examples/bin/run_on_docker.sh
kokosing: do you mean removing docker image? I would not care about it.

kokosing: do you mean removing docker image? I would not care about it. It is like removing files from local maven repositories.

.gitignore
kokosing: now tempto-configuration-local.yaml are committed to repo, so ignoring them in git makes it inconvenient

.travis.yml
kokosing: There are older versions installed in travis. I would prefer to remove this code and use whatever travis has, hence I stick with oldest possible versions of docker and docker-compose. I hope that waiting period for that version in travis should shorter with that approach.

tempto-examples/docker/presto-server/etc/catalog/jmx.properties
kokosing: done

tempto-examples/docker/docker-compose.yml
kokosing: done

tempto-examples/bin/run_on_docker.sh
kokosing: done

tempto-examples/bin/run_on_docker.sh
kokosing: done

tempto-examples/bin/run_on_docker.sh
kokosing: done

tempto-examples/bin/run_on_docker.sh
kokosing: done

tempto-examples/bin/run_on_docker.sh
kokosing: done

tempto-examples/bin/run_on_docker.sh
kokosing: @arhimondr done

tempto-examples/bin/run_on_docker.sh
kokosing: done

tempto-core/src/main/java/com/teradata/tempto/internal/listeners/ProgressLoggingListener.java
losipiuk: whatever. Just 1s seems very arbitrary to me.

tempto-examples/bin/run_on_docker.sh
```diff
@@ -24,55 +24,47 @@ function retry() {
 }
 
 function hadoop_master_container(){
-  docker ps --format '{{.Names}}' | grep hadoop-master
+  ${DOCKER_COMPOSE} ps -q hadoop-master
```
losipiuk: does it work?
I tested with `docker` and it always returned status code 0. That is why I suggested adding `| grep .` at the end of command.

losipiuk:
looks good

losipiuk:
though - travis build failed on this PR. This should probably not happen :)

tempto-core/src/main/java/com/teradata/tempto/internal/listeners/ProgressLoggingListener.java
maciejgrzybek: Can you at least extract this to some constant?

tempto-core/src/test/groovy/com/teradata/tempto/assertions/QueryAssertTest.groovy
```diff
@@ -56,6 +57,11 @@ public class QueryAssertTest
     }
   }
 
+  def setupSpec()
+  {
+    System.setProperty(TEST_CONFIGURATION_URI_KEY, "/configuration/global-configuration-tempto.yaml");
```
maciejgrzybek: Isn't there any better way than introducing config file to tests?
Cannot we just decouple unit tests from config file?

kokosing: It is not just simple. QueryAssert is using static methods it is hacky to use mocking there.

PR #115, opened 2016-04-06
Add use cannonical hostname flag for webhdfs authentication

arhimondr:
On some environments we don't have control over reverse
DNS resolution. So it is worth to have such flag that disables
the reverse DNS resolution.


kokosing:
Code looks ok, but I dont have much context about kerberos

losipiuk:
looks good.

arhimondr:
In order to get the SPNEGO service ticket you must build the service name on client.
By default the hostname canonization is being made via the reverse dns lookup.

Here is the link that can give you more context on that
http://web.mit.edu/Kerberos/krb5-1.13/doc/admin/princ_dns.html#reverse-dns-mismatches

But if the code looks good i will merge that.

PR #116, opened 2016-04-08
Make error message about missing local configuration more neutral

arhimondr:
It is possible to run Tempto only with basic configuration, that is
implemented and packaged into the tests runnable.
We do so in Presto when running product tests from IDE.
It is not that easy to specify local configuration for each test run profile
automatically created by IDE. So we are setting reasonable defaults in default
tempto configuration and letting user to run product tests without local
configuration specified. But the message `WARN: Unable to find local configuration:`
can make user worry that he did something wrong. Instead of displaying such wary
message we decided to display something more neutral.


kokosing:
Lgtm

PR #124, opened 2016-04-25
Add owner function to WebHdfsClient

sanjay990:
Add a function to get the owner of a file/directory on hdfs.
This is required for running kerberos/impersonation product-tests.

### DIFF
--- a/tempto-core/src/main/java/com/teradata/tempto/internal/hadoop/hdfs/WebHdfsClient.java
+++ b/tempto-core/src/main/java/com/teradata/tempto/internal/hadoop/hdfs/WebHdfsClient.java
@@ -183,9 +183,7 @@ public class WebHdfsClient
         }
     }
 
-    @Override
-    @SuppressWarnings("unchecked")
-    public long getLength(String path)
+    private Object getAttributeValue(String path, String attribute)
     {
         HttpGet readRequest = new HttpGet(buildUri(path, "GETFILESTATUS", emptyMap()));
         try (CloseableHttpResponse response = httpRequestsExecutor.execute(readRequest)) {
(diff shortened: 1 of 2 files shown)


tempto-core/src/main/java/com/teradata/tempto/internal/hadoop/hdfs/WebHdfsClient.java
```diff
@@ -194,14 +192,28 @@ public long getLength(String path)
...
         catch (IOException e) {
             throw new RuntimeException("Could not get file status: " + path + " , user: " + username, e);
         }
     }
 
     @Override
+    @SuppressWarnings("unchecked")
+    public long getLength(String path)
+    {
+        return (((Number) getAttributeValue (path, "length")).longValue());
```
cawallin: nit: you don't need the outermost parens

cawallin:
looks good!

tempto-core/src/main/java/com/teradata/tempto/internal/hadoop/hdfs/WebHdfsClient.java
```diff
@@ -183,9 +183,7 @@ public void loadFile(String path, OutputStream outputStream)
         }
     }
 
-    @Override
-    @SuppressWarnings("unchecked")
-    public long getLength(String path)
+    private Object getAttributeValue(String path, String attribute)
```
ebd2: Why not make this public for the benefit of anybody who knows what they're doing and needs to query something that doesn't have a specific method exposed yet?

ebd2:
LGTM

arhimondr:
lgtm

tempto-core/src/main/java/com/teradata/tempto/internal/hadoop/hdfs/WebHdfsClient.java
sanjay990: @ebd2 Yes, we can make it public. But then shouldn't we move this to some other class.
@arhimondr - Any comments.

arhimondr: I think it will be hard to implement this method with using regular HDFS client instead of WebHDFS. We are going to introduce such implementation, because on Facebook clusters WebHDFS is not exposed. So, i would rather keep it private.

arhimondr:
@sanjay990 As i understand it is already merged.

sanjay990:
@arhimondr  - Yes, this is merged.

PR #130, opened 2016-06-16
Treat hive mutable tables as managed tables

kokosing:
Treat hive mutable tables as managed tables


tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/hive/HiveTableManager.java
```diff
@@ -149,29 +136,21 @@ private String getImmutableTableHdfsPath(HiveDataSource dataSource)
...
-        sb.append(testDataBasePath);
-        sb.append(MUTABLE_TABLES_DIR);
-        return sb.toString();
-    }
-
     private String getMutableTableHdfsPath(TableName tableName, Optional<Integer> partitionId)
     {
         StringBuilder sb = new StringBuilder();
-        sb.append(getMutableTablesDir());
+        sb.append("/user/hive/warehouse/");
```
sopel39: extract as static final

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/hive/HiveTableManager.java
```diff
@@ -114,17 +107,11 @@ public HiveTableInstance createMutable(HiveTableDefinition tableDefinition, Stat
             }
         }
         else if (state == LOADED) {
+            String tableDataPath = getMutableTableHdfsPath(tableName, Optional.empty());
```
sopel39: you could assert that the data path doesn't exists (if there isn't such check already)

sopel39:
lgtm

PR #134, opened 2016-07-14
Add more parameters to createViewAs method

amrutagokhale:
This change is needed to write a product test for access control checks in Presto ( These access control checks are turned on when `hive.security=sql-standard` is set ).

### DIFF
--- a/tempto-core/src/main/java/com/teradata/tempto/sql/SqlContexts.java
+++ b/tempto-core/src/main/java/com/teradata/tempto/sql/SqlContexts.java
@@ -14,6 +14,7 @@
 
 package com.teradata.tempto.sql;
 
+import com.teradata.tempto.query.QueryExecutor;
 import com.teradata.tempto.sql.view.ViewContextProvider;
 
 import static com.teradata.tempto.query.QueryExecutor.defaultQueryExecutor;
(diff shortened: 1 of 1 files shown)


tempto-core/src/main/java/com/teradata/tempto/sql/SqlContexts.java
```diff
@@ -41,6 +42,11 @@ public static ViewContextProvider createViewAs(String selectSql)
         return new ViewContextProvider(viewName, selectSql, defaultQueryExecutor());
     }
 
+    public static ViewContextProvider createViewAs(String viewName, String selectSql, QueryExecutor queryExecutor)
+    {
+        return new ViewContextProvider(viewName, selectSql, queryExecutor);
```
akshatnair: Why can't you use defaultQueryExecutor? Your commit message should ideally tell why something is being done rather than what is being done.

amrutagokhale: Ok, I will updated the commit message. I want to use a custom QueryExecutor in which I am creating a custom connection with a different user (say `alice`) than the default `hdfs` user. This is needed for the access control checks tests (for granting/revoking privileges, we need more than a single user).

amrutagokhale: Updated the commit message.

akshatnair: Cool, thanks. LGTM. I will post this in the testing framework chat so that someone else can look at it as well.

amrutagokhale: Ok, thanks. I will wait for one more review.

brian-rickman:
LGTM

PR #136, opened 2016-07-27
Implement LDAP Object creation requirement

anusudarsan:
Added a new requirement and fulfiller for creating LDAP Objects.
The tempto config required are the LDAP Server url and the admin
credentials for the LDAP server.

The requirement takes a list of LDAP Object definitions and will be
executed in that order. If the LDAP Entry already exists, then the addition is ignored.


anusudarsan:
@losipiuk can you please review?

README.md
```diff
@@ -431,6 +431,58 @@ One can request that mutable table is in one of three states:
...
+
+```ldap:
+     url: ldap://ldapserverhost
+     admin:
+       dn: cn=admin,dc=tempto,dc=com
+       password: admin
+```
+
+The LDAP Server used must be configured with an admin user and password who has privileges to create entities.
+If a similar entry already exists in the LDAP server, then the fulfiller will ignore adding that entry.
```
losipiuk: To what entry does this line relate? It seems to me that this not should be part of `LdapObjectDefinition` documentation.

tempto-core/src/main/java/com/teradata/tempto/fulfillment/ldap/LdapObjectDefinition.java
```diff
@@ -0,0 +1,98 @@
...
+
+    public static class LdapObjectDefinitionBuilder
+    {
+
+        private String id;
+        private String distinguishedName;
+        private List<String> objectClasses;
+        private Map<String, String> attributes;
+
+        public LdapObjectDefinitionBuilder(String id)
```
losipiuk: make this one non-public to enforce useage `.builder` static method.

tempto-core/src/main/java/com/teradata/tempto/fulfillment/ldap/LdapObjectDefinition.java
```diff
@@ -0,0 +1,98 @@
...
+    private final String id;
+    private final String distinguishedName;
+    private final Map<String, String> attributes;
+    private final List<String> objectClasses;
+
+    public LdapObjectDefinition(String id, String distinguishedName, Map<String, String> attributes, List<String> objectClasses)
+    {
+        this.id = requireNonNull(id, "id is null");
+        this.distinguishedName = requireNonNull(distinguishedName, "distinguishedName is null");
+        this.attributes = requireNonNull(attributes, "attributes is null");
```
losipiuk: make map copy

tempto-core/src/main/java/com/teradata/tempto/fulfillment/ldap/LdapObjectDefinition.java
```diff
@@ -0,0 +1,98 @@
...
+        this.objectClasses = requireNonNull(objectClasses, "objectClasses is null");
```
losipiuk: make list copy

tempto-core/src/main/java/com/teradata/tempto/fulfillment/ldap/LdapObjectDefinition.java
```diff
@@ -0,0 +1,98 @@
...
+
+        public LdapObjectDefinitionBuilder setDistinguishedName(String distinguishedName)
+        {
+            this.distinguishedName = distinguishedName;
+            return this;
+        }
+
+        public LdapObjectDefinitionBuilder setAttributes(Map<String, String> attributes)
+        {
+            this.attributes = attributes;
```
losipiuk: make map copy

tempto-core/src/main/java/com/teradata/tempto/fulfillment/ldap/LdapObjectDefinition.java
```diff
@@ -0,0 +1,98 @@
...
+            return this;
+        }
+
+        public LdapObjectDefinitionBuilder addObjectClasses(List<String> objectClasses)
+        {
+            this.objectClasses = objectClasses;
```
losipiuk: make list copy

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/ldap/DefaultLdapObjectEntryManager.java
```diff
@@ -0,0 +1,131 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+import static java.lang.String.format;
+import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.toList;
+
+public class DefaultLdapObjectEntryManager
+        implements LdapObjectEntryManager
+{
+    private final DirContext context;
+    private static final Logger LOGGER = LoggerFactory.getLogger(DefaultLdapObjectEntryManager.class);
```
losipiuk: move LOGGER above context.

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/ldap/DefaultLdapObjectEntryManager.java
```diff
@@ -0,0 +1,131 @@
...
+
+    @Inject
+    public DefaultLdapObjectEntryManager(
```
losipiuk: Wouldn't `OpenLdapObjectEntryManager` be better name for this one?

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/ldap/LdapObjectEntryManager.java
```diff
@@ -0,0 +1,24 @@
...
+
+package com.teradata.tempto.internal.fulfillment.ldap;
+
+import com.teradata.tempto.fulfillment.ldap.LdapObjectDefinition;
+
+import java.util.List;
+
+interface LdapObjectEntryManager
+{
+    List<String> addLdapDefinitions(List<LdapObjectDefinition> ldapObjectDefinitions);
```
losipiuk: Don't we need to clean those entries at the end of tests run?

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/ldap/LdapObjectModuleProvider.java
```diff
@@ -0,0 +1,39 @@
...
+package com.teradata.tempto.internal.fulfillment.ldap;
+
+import com.google.inject.AbstractModule;
+import com.google.inject.Module;
+import com.google.inject.Scopes;
+import com.teradata.tempto.configuration.Configuration;
+import com.teradata.tempto.initialization.AutoModuleProvider;
+import com.teradata.tempto.initialization.SuiteModuleProvider;
+
+@AutoModuleProvider
```
losipiuk: With this in tempto-core jar we will always need `ldap` section in config even if product tests are not using ldap at all.
Othwerwise we will fail on setup.
Short term solution to that is to create separate maven module `tempto-ldap` and only explicitly add dependency to `tempto-ldap.jar` for product tests projects which require that.

Long there I think we should get rid of that `Auto*` stuff and require users to list fulfillers explicitly during tempto initialization.

For now move ldap stuff to separate module.

tempto-examples/src/main/resources/tempto-configuration.yaml
```diff
@@ -84,3 +84,9 @@ list:
     - ala
     - ma
     - kota
+
+ldap:
```
losipiuk: would be nice to add some example (which to some extend serve as tests for tempto) which uses ldap.
This is some work to be done though as docker machine with ldap would be needed. Decide if it is worth the work.

README.md
anusudarsan: right. updated

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/ldap/DefaultLdapObjectEntryManager.java
anusudarsan: This does not have anything OpenLDAP specific, and should ideally work for say, Active directory too. (I added a note in the README that we support only OpenLDAP because we are planning to test this only using OpenLDAP).

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/ldap/LdapObjectEntryManager.java
anusudarsan: If an entry already exist, we just ignore it for now. We want to actually get some tests running with this, and if its feel necessary I can add it in a separate patch.

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/ldap/LdapObjectModuleProvider.java
anusudarsan: done.

tempto-examples/src/main/resources/tempto-configuration.yaml
anusudarsan: Yes, we are planning to get OpenLDAP on docker next sprint. Once we have that I can add some example tests in a new patch. Do you want me to revert the change in this file?

anusudarsan:
@losipiuk addressed comments

tempto-core/src/main/java/com/teradata/tempto/fulfillment/ldap/LdapObjectDefinition.java
losipiuk: I mean the constructor for builder. But actually both should be private.

tempto-ldap/src/main/java/com/teradata/tempto/fulfillment/ldap/LdapObjectDefinition.java
```diff
@@ -0,0 +1,101 @@
...
+
+    public static class LdapObjectDefinitionBuilder
+    {
+
+        private String id;
+        private String distinguishedName;
+        private List<String> objectClasses;
+        private Map<String, String> attributes;
+
+        public LdapObjectDefinitionBuilder(String id)
```
losipiuk: I meant this constructor should be private. You made constructor of `LdapObjectDefinition` private. Actually both should be.

tempto-examples/src/main/resources/tempto-configuration.yaml
```diff
@@ -84,3 +84,9 @@ list:
...
+  url: ldap://ldapserverhost
```
losipiuk: Now when ldap support is in separate jar (`tempto-ldap`) add it as dependency of tempto-examples so this part of config is used.

losipiuk:
Two small comments. Otherwise looks good to me.

anusudarsan:
Addressed and merged. thanks!

PR #143, opened 2016-08-05
Use DECIMAL(12,2) instead of DOUBLE

kokosing:
Use DECIMAL(12,2) instead of DOUBLE

Make TPC-H tables to be compliant with TPC-H specification by using
DECIMAL(12,2) instead of DOUBLE.

Spec:
```
Decimal means that the column must be able to represent values in the
range -9,999,999,999.99 to
+9,999,999,999.99 in increments of 0.01; the values can be either
represented exactly or interpreted to be in this range;
```


sopel39:
Does querying such table work, e.g: select * from part?
In textfile such decimals would have scale larger then 2 I think

kokosing:
It works. We could have tests for that in case of reading DECIMAL(10,2) from `23.2134123` should round it correctly.

pnowojski:
lgtm if it doesn't brake anything

PR #146, opened 2016-08-10
Not fail when hive database path does not end with /

### LINKED ISSUE #140: Convention based hive mutable tables are unable long when hive.database.path does not end with `/`

kokosing:
Not fail when hive database path does not end with /

Fixes #140

Reviewers: ksobczak


sopel39:
lgtm

PR #148, opened 2016-08-10
Open connection to db only when queries are executed

### LINKED ISSUE #142: Lazy initialization of databases from configuration.yaml
When there are no tests scheduled for execution using database  configured in yaml configuration, the database should not be connected.
E.g.:
tempto yaml:
```yaml
databases:
  default:
    alias: presto

  hive:
    host: hadoop-master

  mysql:
    jdbc_driver_class: com.mysql.jdbc.Driver
    jdbc_url: jdbc:mysql://mysql:3306/test
    jdbc_user: root
```

running only Hive tests (e.g. `TestAllDatatypesFromHiveConnector.java`) MySQL instance shouldn't be expected to run.

kokosing:
Open connection to db only when queries are executed

Fixes #142


tempto-examples/src/main/resources/tempto-configuration-wrong.yaml
```diff
@@ -0,0 +1,84 @@
+# This configuration contains only presto and hive configured correctly
```
losipiuk: Give it more meaningful name.
Genrally we want to allow only specific misconfigurations in yaml file. 
And this tests suggests that any errors are good.

PR #153, opened 2016-08-25
Raise an exception when no tests executed

kokosing:
Raise an exception when no tests executed


ArturGajowy:
I think this is a sane default, but should come with an option to allow no tests executed, e.g. `-DfailOnNoTest=false`. Ohterwise this will wreck people's git rebase -x / git bisect workflows.

ArturGajowy:
Ohterwsie LGTM :)

kokosing:
thanks

PR #176, opened 2016-10-17
Move Table repository to Runner package

maciejgrzybek:
This is a prerequisite to extract hive and jdbc into separate modules.


tempto-examples/src/main/java/com/teradata/tempto/examples/SimpleQueryTest.java
```diff
@@ -38,8 +38,8 @@
 import static com.teradata.tempto.context.ThreadLocalTestContextHolder.testContext;
 import static com.teradata.tempto.context.ThreadLocalTestContextHolder.testContextIfSet;
 import static com.teradata.tempto.fulfillment.table.hive.HiveTableDefinition.like;
-import static com.teradata.tempto.fulfillment.table.hive.tpch.TpchTableDefinitions.NATION;
-import static com.teradata.tempto.fulfillment.table.hive.tpch.TpchTableDefinitions.REGION;
+import static com.teradata.tempto.runner.fulfillment.table.hive.tpch.TpchTableDefinitions.NATION;
```
cawallin: Will this require changes for the presto/qg product tests?

maciejgrzybek: Yes, it will.

maciejgrzybek:
Unfortunately, we cannot avoid changing package if we want to introduce modularity. Currently Hive and JDBC both depend on tempto-core implicitly. Both definition and running of tests stay in the same module, thus it's not possible to introduce new module (namely tempto-cassandra) to reuse some parts of JDBC codebase because it would introduce circular dependency (cassandra depending on JDBC (which is currently tempto-core) and tempto-core depending on cassandra (to execute cassandra-type tests)).

cawallin:
OK, that's fine. looks good!

arhimondr:
lgtm

tempto-runner/src/main/resources/log4j.properties
```diff
@@ -0,0 +1,29 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# Set root logger to INFO and add a custom appender.
+log4j.rootLogger=TRACE, TEST_FRAMEWORK_LOGGING_APPENDER, CONSOLE
```
kokosing: this file should go in separate commit. 

AFAIK by default tempto-runner is not bound to any of logging providers, so I see no reason why this file should be commited.

I can be used in tempto-example, but presto does not use log4j.

tempto-core/src/main/java/com/teradata/tempto/fulfillment/table/RepositoryTableDefinition.java
```diff
@@ -0,0 +1,29 @@
+/*
```
kokosing: I don't understand why do you move this class to runner.

If you need to create modules for hive and jdbc, then I could imagine that hive and jdbc are using core, and runner is using core, hive and jdbc. Why do you put it here?

maciejgrzybek: That's because of `ConventionTableDefinitionsProvider` is familiar with all modules we have (have a look on method `tableDefinitionFor`).
`ConventionTableDefinitionsProvider` is used by `TableDefinitionsRepository` which uses `RepositoryTableDefinition`.
If we extract `hive` and `tpch` to external module (outside core) and keep `RepositoryTableDefinition` (and classes mentioned above) we'd end up with circular dependency (core -> hive (because of `ConventionTableDefinitionsProvider`'s `tableDefinitionFor` and hive -> core)).

kokosing:
Maybe we could have table repository in core module? It would require tpch
to be Independent from hive, but it would be cleaner. Runner should be only
responsible for running testing and parsing click options
Usage of runner should be optional to run tempto tests.

wt., 18.10.2016, 15:39 użytkownik maciejgrzybek <notifications@github.com>
napisał:

kokosing:
And what is the motivation behind extracting tpch, jdbc and hive module?

wt., 18.10.2016, 18:24 użytkownik Grzegorz Kokosiński <
g.kokosinski@gmail.com> napisał:

maciejgrzybek:
AD https://github.com/prestodb/tempto/pull/176#issuecomment-254561964: How would you implement `ConventionTableDefinitionsProvider` inside tempto-core having `hive` as an independent module?

AD https://github.com/prestodb/tempto/pull/176#issuecomment-254564343: The rationale is one may not be interested in testing Hive but testing Postgres only. Why would he bother having all deps installed? And obviously maintenance would be easier.
I claim HDFS shouldn't be mandatory part of framework.
I CC @losipiuk because he stated tempto-cassandra, I'm working on, should be a separate module (I agree with that statement).

losipiuk:
> AD #176 (comment): How would you implement ConventionTableDefinitionsProvider inside tempto-core having hive as an independent module?

I talked with @kokosing and I do not think that tempto-runner should have any mandatory classes which are needed by tempto at runtime.
The module graph should look like @kokosing pointed before

```
         tempto-hive +---------+
                               |
                               |
         tempto-jdbc +-----------------> tempto-core
                               |
                               |              ^
         tempto-cassandra +----+              |
                                              |
                                              |
                                              |
                                              |
         tempto-runner +----------------------+
```

And the it relates to application using presto in such way:
```
                                  +- - - - - - -> tempto-runner
                                  |
                                  |
                                  +-------------> tempto-core
                                  |
                                  |
    application-using-tempto +------ - - - - - -> tempto-hive
                                  |
                                  |
                                  +- - - - - - -> tempto-jdbc
                                  |
                                  |
                                  +- - - - - - -> tempto-cassandra


                                    +- - - - - - + optional

```

To get rid of compile time dependencies from `tempto-core` to `tempto-*` we should use classpath scanning. It is already happening right now. See how fullfillers/table managers are registered.
This can be extended to handle convention based tables.
Generally `ConventionTableDefinitionsProvider` would need to be changed to be parametrized with list of lets-call-them `SubConventionTableDefinitionsProvider`. And list of `SubConventionTableDefinitionsProvider` would be gathered using classpath scan. 

Ideally we should introduce some `Plugin` interface like Presto have. `Plugin` interface would serve as single extension point for module. 
And expose all the fullfillers/table_managers/sub_convention_table_definition_providers.
But this is strictly optional followup refactoring.

On side node.
Generally I am not big fan of classpath scanning. And rather would require user to list all the requested modules explicitly in applictaion-using-presto. But then the right-click-run-test in IntelliJ would not work. And I bet we want that.

maciejgrzybek:
@losipiuk we came to the very similar conclusion in offline discussion with @arhimondr.
If you guys don't mind I'd prefer explicit registration of table managers / data providers (everything you called "plugin" above) instead of class path scanning.
The point is it's much easier to debug if you explicitly register classes compared to struggle with automagically leaking classes onto class path and being loaded implicitly.
If you don't object I'd rewrite it in such way as @losipiuk proposed modulo explicit registration of plugins (e.g. registry would be in tempto-core but tempto-runner would have pass-through also, to make registration easier in application).

losipiuk:
If we want to sacrifice running tests without Runner in IntelliJ - then OK.
I personally like that feature and use it quite a lot.
Or maybe you have an idea which will still keep IntelliJ running working?

ArturGajowy:
@maciejgrzybek if making sense of classpath-added plugins is a problem, please add a diagnostic log telling which plugins were loaded. I don't think we'll ever have a case where we won't be able to change the classpath to our needs, so I'm against sacrificing IntelliJ runnability.

maciejgrzybek:
Superseded by: https://github.com/prestodb/tempto/pull/179

PR #177, opened 2016-10-18
Do not create directory in HDFS when creating ImmutableTablesFulfiller

kokosing:
Do not create directory in HDFS when creating ImmutableTablesFulfiller

When creating ImmutableTablesFulfiller, it creates TableManagers, and so
creates HiveTableManager, which then creates
DefaultHdfsDataSourceWriter. This needs RevisionStorage to be created
first. In order to create RevisionStorage a DefaultRevisionProvider,
creates a directory in HDFS to decide which RevisionStorage it should
create.

Above causes a problem when running tests which does not require hdfs to
be provisioned.

By using Provider explicitly, RevisionStorage is created just before any
write or read to or from HDFS,


tempto-core/src/main/java/com/teradata/tempto/internal/hadoop/hdfs/revisions/DispatchingRevisionStorage.java
```diff
@@ -0,0 +1,121 @@
...
+
+    private static final String TEST_X_ATTR_KEY = "user.test-attr-key";
+    private static final String TEST_X_ATTR_VALUE = "test-attr-value";
+
+    private final Provider<RevisionStorage> revisionStorage;
+
+    @Inject
+    public DispatchingRevisionStorage(HdfsClient hdfsClient, @Named(CONF_TESTS_HDFS_PATH_KEY) String testDataBasePath)
+    {
+        revisionStorage = new CachingProvider(new HdfsRevisionStorageProvider(hdfsClient, testDataBasePath));
```
losipiuk: Name it LazyCachingProvider

tempto-core/src/main/java/com/teradata/tempto/util/CachingProvider.java
```diff
@@ -0,0 +1,40 @@
...
+    private final Provider<T> provider;
+    private T instance;
+
+    public CachingProvider(Provider<T> provider)
+    {
+        this.provider = requireNonNull(provider, "provider is null");
+    }
+
+    @Override
+    public T get()
```
losipiuk: make it synchronized

PR #179, opened 2016-10-21
Create a stub of SPI for testing modules

maciejgrzybek:
This enables us to remove hack in `ConventionTableDefinitionsProvider` with a switch containing all existsing "plugins".
This patch is a prerequisite to extract Hive and Jdbc plugins to separate modules.

### DIFF
--- a/tempto-core/src/main/java/com/teradata/tempto/fulfillment/table/TableDefinitionsRepository.java
+++ b/tempto-core/src/main/java/com/teradata/tempto/fulfillment/table/TableDefinitionsRepository.java
@@ -13,9 +13,12 @@
  */
 package com.teradata.tempto.fulfillment.table;
 
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.MapMaker;
 import com.teradata.tempto.internal.convention.tabledefinitions.ConventionTableDefinitionsProvider;
 import com.teradata.tempto.internal.ReflectionHelper;
+import com.teradata.tempto.spi.Plugin;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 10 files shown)


tempto-core/src/main/java/com/teradata/tempto/hive/HivePlugin.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.Optional;
+
+import static com.teradata.tempto.fulfillment.table.TableHandle.tableHandle;
+import static com.teradata.tempto.fulfillment.table.hive.HiveTableDefinition.hiveTableDefinition;
+
+public class HivePlugin
+    implements Plugin
+{
+    @Override
+    public TableDefinition getConventionTableDefinition(ConventionTableDefinitionDescriptor descriptor)
```
losipiuk: I would rather have `Plugin` expose sth like `ConventionTableDefinitionProvider`.
And then provider would have method `getConventionTableDefinition`. 
We do not want to pass `Plugin` interface all around the codebase to places when only subset of things exposed by it are needed.

tempto-core/src/main/java/com/teradata/tempto/fulfillment/table/TableDefinitionsRepository.java
```diff
@@ -62,9 +71,26 @@
...
+                try {
+                    Constructor<? extends Plugin> constructor = plugin.getConstructor();
+                    Plugin p = constructor.newInstance();
+                    builder.put(p.getName(), p);
+                }
+                catch (NoSuchMethodException e) {
+                    LOGGER.warn("Plugin %s does not have appropriate constructor. Skipping", plugin);
+                }
+                catch (IllegalAccessException|InstantiationException|InvocationTargetException e) {
+                    LOGGER.warn("Cannot instantiate plugin %s: %s. Skipping.", plugin, e);
```
kokosing: also add a `LOGGER.info` that plugin was loaded

tempto-core/src/main/java/com/teradata/tempto/hive/HivePlugin.java
```diff
@@ -0,0 +1,43 @@
+package com.teradata.tempto.hive;
```
kokosing: copyrights

kokosing:
@maciejgrzybek Are you going to work on this? Or should i close it?

maciejgrzybek:
We need to rethink and possibly revamp the testing infra at some point. 
It does not make sense to invest time for this PR now, especially given the solution I proposed bears some difficulties in implementation which may not be straightforward to solve.

Closing for now.
CC: @ebd2

PR #180, opened 2016-10-27
Add Cassandra fulfillers

maciejgrzybek:
Please disregard commit msgs and their order. This is just a gist of implementation.

### DIFF
--- a/tempto-core/src/main/java/com/teradata/tempto/fulfillment/table/jdbc/RelationalDataSource.java
+++ b/tempto-core/src/main/java/com/teradata/tempto/fulfillment/table/jdbc/RelationalDataSource.java
@@ -18,7 +18,7 @@ import java.sql.PreparedStatement;
 import java.util.Iterator;
 import java.util.List;
 
-public interface JdbcTableDataSource
+public interface RelationalDataSource
 {
     /**
      * Returns iterator over rows to be inserted to table.
--- a/tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/TableManagerDispatcherModuleProvider.java
+++ b/tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/TableManagerDispatcherModuleProvider.java
@@ -86,8 +86,11 @@ public class TableManagerDispatcherModuleProvider
                     @Override
                     protected void configure()
                     {
-                        // we bind matching QueryExecutor to be visible by TableManager without @Named annotation
-                        bind(QueryExecutor.class).to(Key.get(QueryExecutor.class, named(database)));
+                        // TODO: refactor this to avoid hardcoding dbname
+                        if (!database.equals("cassandra")) { // Cassandra does not provide QueryExecutor
+                            // we bind matching QueryExecutor to be visible by TableManager without @Named annotation
+                            bind(QueryExecutor.class).to(Key.get(QueryExecutor.class, named(database)));
+                        }
                         bind(Key.get(String.class, named("databaseName"))).toInstance(database);
                         bind(tableManagerKey).to(tableManagerClass);
                         expose(tableManagerKey);
--- a/tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/TableName.java
+++ b/tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/TableName.java
@@ -53,6 +53,11 @@ public class TableName
         return nameInDatabase;
     }
 
+    public String getSchemalessNameInDatabase()
+    {
+        return nameInDatabase;
+    }
+
     public Optional<String> getSchema()
     {
         return schema;
--- a/tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraBatchLoader.java
+++ b/tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraBatchLoader.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.teradata.tempto.internal.fulfillment.table.cassandra;
+
+import com.datastax.driver.core.BatchStatement;
+import com.datastax.driver.core.PreparedStatement;
+import com.datastax.driver.core.Session;
+
+import java.sql.SQLException;
+import java.util.List;
+
+import static java.lang.String.format;
+import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.joining;
+
+public class CassandraBatchLoader
+{
+    private final Session session;
+    private final String tableName;
+    private final List<String> columnNames;
+
+    public CassandraBatchLoader(Session session, String tableName, List<String> columnNames)
+    {
+        this.session = requireNonNull(session, "session is null");
+        this.tableName = requireNonNull(tableName, "tableName is null");
+        this.columnNames = requireNonNull(columnNames, "columnNames is null");
+    }
+
+    public void load(List<List<Object>> batch)
+            throws SQLException
+    {
+        PreparedStatement statement = session.prepare(format("INSERT INTO %s (%s) VALUES(%s)",
+                tableName,
+                columnNames.stream().collect(joining(",")),
+                repeatPattern("?", ",", columnNames.size())));
+
+        BatchStatement batchStatement = new BatchStatement(BatchStatement.Type.UNLOGGED);
+
+        for (List<Object> row : batch) {
+            batchStatement.add(statement.bind(row.toArray()));
+        }
+
+        session.execute(batchStatement);
+    }
+
+    private String repeatPattern(String ch, String separator, int times)
+    {
+        String result = "";
+
+        for (int i = 0; i < times - 1; ++i) {
+            result += ch + separator;
+        }
+
+        return result + ch;
+    }
+}
--- a/tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraTableManager.java
+++ b/tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraTableManager.java
@@ -1,0 +1,229 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.teradata.tempto.internal.fulfillment.table.cassandra;
+
+import com.google.inject.Inject;
+import com.teradata.tempto.configuration.Configuration;
+import com.teradata.tempto.fulfillment.table.MutableTableRequirement;
+import com.teradata.tempto.fulfillment.table.TableDefinition;
+import com.teradata.tempto.fulfillment.table.TableHandle;
+import com.teradata.tempto.fulfillment.table.TableInstance;
+import com.teradata.tempto.fulfillment.table.TableManager;
+import com.teradata.tempto.fulfillment.table.jdbc.RelationalDataSource;
+import com.teradata.tempto.fulfillment.table.jdbc.RelationalTableDefinition;
+import com.teradata.tempto.internal.fulfillment.table.TableName;
+import com.teradata.tempto.internal.fulfillment.table.TableNameGenerator;
+import com.teradata.tempto.internal.query.CassandraQueryExecutor;
+import org.slf4j.Logger;
+
+import javax.inject.Named;
+import javax.inject.Singleton;
+
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkState;
+import static com.teradata.tempto.fulfillment.table.MutableTableRequirement.State.CREATED;
+import static com.teradata.tempto.fulfillment.table.MutableTableRequirement.State.LOADED;
+import static com.teradata.tempto.fulfillment.table.MutableTableRequirement.State.PREPARED;
+import static java.lang.String.format;
+import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.toList;
+import static org.slf4j.LoggerFactory.getLogger;
+
+@TableManager.Descriptor(tableDefinitionClass = RelationalTableDefinition.class, type = "CASSANDRA")
+@Singleton
+public class CassandraTableManager
+        implements TableManager<RelationalTableDefinition>
+{
+    private static final Logger LOGGER = getLogger(CassandraTableManager.class);
+
+    private static final int BATCH_SIZE = 10_000;
+
+    private final TableNameGenerator tableNameGenerator;
+    private final CassandraQueryExecutor queryExecutor;
+    private final String databaseName;
+    private final String defaultKeySpace;
+
+    @Inject
+    public CassandraTableManager(
+            TableNameGenerator tableNameGenerator,
+            @Named("databaseName") String databaseName,
+            Configuration configuration)
+    {
+        this.tableNameGenerator = requireNonNull(tableNameGenerator, "tableNameGenerator is null");
+        this.queryExecutor = new CassandraQueryExecutor(configuration);
+        this.databaseName = requireNonNull(databaseName, "databaseName is null");
+        this.defaultKeySpace = configuration.getStringMandatory("databases.cassandra.default_schema");
+    }
+
+    @Override
+    public TableInstance<RelationalTableDefinition> createImmutable(RelationalTableDefinition tableDefinition, TableHandle tableHandle)
+    {
+        TableName tableName = createImmutableTableName(tableHandle);
+        checkArgument(tableName.getSchema().isPresent(), "Cassandra requires table with schema");
+
+        if (!queryExecutor.tableExists(tableName.getSchema().get(), tableName.getSchemalessNameInDatabase())) {
+            executeQueryIgnoreTypeError(tableDefinition.getCreateTableDDL(tableName.getNameInDatabase()));
+            RelationalDataSource dataSource = tableDefinition.getDataSource();
+
+            // TODO: check if data is up-to-date
+
+            insertData(tableName, dataSource);
+        }
+
+        return new CassandraTableInstance(tableName, tableDefinition);
+    }
+
+    private void insertData(TableName tableName, RelationalDataSource dataSource)
+    {
+        checkArgument(tableName.getSchema().isPresent(), "Cassandra requires table with schema");
+
+        Iterator<List<Object>> dataRows = dataSource.getDataRows();
+        if (!dataRows.hasNext()) {
+            return;
+        }
+
+        List<String> columnNames = queryExecutor.getColumnNames(tableName.getSchema().get(), tableName.getSchemalessNameInDatabase());
+
+        try {
+            CassandraBatchLoader loader = new CassandraBatchLoader(queryExecutor.getSession(), tableName.getNameInDatabase(), columnNames);
+            for (List<List<Object>> batch : partitionBy(dataRows, BATCH_SIZE)) {
+                loader.load(batch);
+            }
+        }
+        catch (SQLException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private static Iterable<List<List<Object>>> partitionBy(Iterator<List<Object>> dataRows, int partitionSize)
+    {
+        return () -> new Iterator<List<List<Object>>>()
+        {
+            @Override
+            public boolean hasNext()
+            {
+                return dataRows.hasNext();
+            }
+
+            @Override
+            public List<List<Object>> next()
+            {
+                List<List<Object>> batch = new ArrayList<>();
+                while (dataRows.hasNext() && batch.size() < partitionSize) {
+                    batch.add(dataRows.next());
+                }
+                return batch;
+            }
+        };
+    }
+
+    @Override
+    public TableInstance<RelationalTableDefinition> createMutable(RelationalTableDefinition tableDefinition, MutableTableRequirement.State state, TableHandle tableHandle)
+    {
+        TableName tableName = createMutableTableName(tableHandle);
+        checkArgument(tableName.getSchema().isPresent(), "Cassandra requires table with schema");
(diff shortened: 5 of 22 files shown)


tempto-core/src/main/java/com/teradata/tempto/fulfillment/table/jdbc/RelationalDataSource.java
```diff
@@ -18,7 +18,7 @@
 import java.util.Iterator;
 import java.util.List;
 
-public interface JdbcTableDataSource
+public interface RelationalDataSource
```
kokosing: This is somehow wierd. `HiveTableDefinition` can be also though as `relational`, in that sense this name is wrong.

If cassandra does not support JDBC maybe we would need some special `CassandraTableDefinition` as we have for hive. It think that way seems more natural.

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraBatchLoader.java
```diff
@@ -0,0 +1,65 @@
+package com.teradata.tempto.internal.fulfillment.table.cassandra;
```
kokosing: this missing licence causes travis to fail

tempto-core/src/main/java/com/teradata/tempto/internal/query/CassandraQueryExecutor.java
```diff
@@ -0,0 +1,151 @@
...
+        builder.put(DataType.cfloat(), JDBCType.REAL);
+        builder.put(DataType.cint(), JDBCType.INTEGER);
+        builder.put(DataType.smallint(), JDBCType.SMALLINT);
+//        builder.put(DataType.text(), JDBCType.NVARCHAR);
+        builder.put(DataType.time(), JDBCType.TIME);
+        builder.put(DataType.timestamp(), JDBCType.TIMESTAMP);
+        builder.put(DataType.tinyint(), JDBCType.TINYINT);
+        builder.put(DataType.varchar(), JDBCType.VARCHAR);
+
+        typeMapping = builder.build();
```
kokosing: use chaining

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/TableManagerDispatcherModuleProvider.java
```diff
@@ -86,8 +86,10 @@ protected void configure()
                     @Override
                     protected void configure()
                     {
-                        // we bind matching QueryExecutor to be visible by TableManager without @Named annotation
-                        bind(QueryExecutor.class).to(Key.get(QueryExecutor.class, named(database)));
+                        if (!database.equals("cassandra")) {
```
kokosing: I think this `if` requires at least a comment.

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/TableName.java
```diff
@@ -53,6 +53,11 @@ public String getNameInDatabase()
         return nameInDatabase;
     }
 
+    public String getInternalNameInDatabase()
```
kokosing: how does it differ to `getNameInDatabase`?

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraTableManager.java
```diff
@@ -52,13 +52,16 @@ public CassandraTableManager(TableNameGenerator tableNameGenerator,
     public TableInstance<RelationalTableDefinition> createImmutable(RelationalTableDefinition tableDefinition, TableHandle tableHandle)
     {
         TableName tableName = createImmutableTableName(tableHandle);
-        if (!tableName.getSchema().isPresent()) {
-            // drop table
-        }
-         QueryResult rs = queryExecutor.executeQuery(tableDefinition.getCreateTableDDL(tableName.getNameInDatabase()));
+        checkArgument(tableName.getSchema().isPresent(), "Cassandra does not like you when you don't have keyspace defined");
```
kokosing: nice message

tempto-core/src/main/java/com/teradata/tempto/fulfillment/table/jdbc/RelationalDataSource.java
maciejgrzybek: My idea here was to use "Relational" as a name here since data is produced as tuples for relation (row by row). Our `HiveDataSource` loads data in binary format (e.g. reading from resources and uploading to HDFS). I believe `HiveDataSource` should be renamed to `BinaryDataSource` % a few changes to the interface.
It's a work for a separate PR though.

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/TableName.java
maciejgrzybek: It returns name of table without schema.
Cassandra driver requires schema to given separately.

tempto-core/src/main/java/com/teradata/tempto/internal/query/CassandraQueryExecutor.java
maciejgrzybek: 16fbcd24205ae5facbeedfc5d87656602f66c8f8

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraTableManager.java
```diff
@@ -0,0 +1,209 @@
...
+        return tableInstance;
+    }
+
+    @Override
+    public void dropTable(TableName tableName)
+    {
+        queryExecutor.executeQuery("DROP TABLE " + tableName.getNameInDatabase());
+    }
+
+    public void dropTable(String tableName)
```
maciejgrzybek: One `dropTable` (implementation of the interface method) is public, so I think for consistency it would be good to have this public as well. I can change that if you want though.

cawallin: It seems a little different b/c the other one is marked with  `@Override`, but I don't feel strongly

tempto-core/src/main/java/com/teradata/tempto/internal/query/CassandraQueryExecutor.java
```diff
@@ -0,0 +1,155 @@
...
+                .put(DataType.blob(), JDBCType.BLOB)
+                .put(DataType.cboolean(), JDBCType.BOOLEAN)
+                .put(DataType.counter(), JDBCType.BIGINT)
+                .put(DataType.date(), JDBCType.DATE)
+                .put(DataType.decimal(), JDBCType.DECIMAL)
+                .put(DataType.cdouble(), JDBCType.DOUBLE)
+                .put(DataType.cfloat(), JDBCType.REAL)
+                .put(DataType.cint(), JDBCType.INTEGER)
+                .put(DataType.smallint(), JDBCType.SMALLINT)
+                //.put(DataType.text(), JDBCType.NVARCHAR)
```
maciejgrzybek: There's no SET/MAP/LIST type in JDBC. Presto treats them as "clob"/"varchar" but it's not universal, so I believe we shouldn't add them unless we need them.
Seems like there's no universal method to convert Cassandra types to JDBC types. In Presto's `CassandraMetadata` we have mapping hardcoded.
I don't believe we should blindly follow how we map types there because e.g. MAP -> VARCHAR mapping is valid for Presto only. It's not universally correct.

Moreover, I don't think this mapping is super important now because the only use of `CassandraQueryExecutor.executeQuery` is to create Cassandra tables. So the output (where JDBC translation happens) is not used at all for now.

When we notice we need to expose `executeQuery` to the test case (e.g. expose `executeOnCassandra` method), we'll have more issues than just how to map non-standard types.

I vote for postponing resolution of this problem.

tempto-core/src/main/java/com/teradata/tempto/internal/query/CassandraQueryExecutor.java
```diff
@@ -0,0 +1,155 @@
...
+                .put(DataType.time(), JDBCType.TIME)
+                .put(DataType.timestamp(), JDBCType.TIMESTAMP)
+                .put(DataType.tinyint(), JDBCType.TINYINT)
+                .put(DataType.varchar(), JDBCType.VARCHAR)
+                .build();
+    }
+
+    @Inject
```
maciejgrzybek: `Configuration` is injected.

cawallin: But don't you create it directly?

tempto-core/src/main/java/com/teradata/tempto/internal/query/CassandraQueryExecutor.java
```diff
@@ -0,0 +1,155 @@
...
+        KeyspaceMetadata keyspaceMetadata = cluster.getMetadata().getKeyspace(keySpace);
+        if (keyspaceMetadata == null) {
+            return false;
+        }
+        return keyspaceMetadata.getTable(tableName) != null;
+    }
+
+    public List<String> getTableNames(String keySpace)
+    {
+        return cluster.getMetadata().getKeyspace(keySpace).getTables().stream().map(TableMetadata::getName).collect(toList());
```
maciejgrzybek: ff98c15b97b23395729cd7ecbcd193a46f50e6db

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraTableManager.java
```diff
@@ -0,0 +1,209 @@
...
+    @Override
+    public Class<? extends TableDefinition> getTableDefinitionClass()
+    {
+        return RelationalTableDefinition.class;
+    }
+
+    @Override
+    public void dropStaleMutableTables()
+    {
+        List<String> tableNames = queryExecutor.getTableNames(defaultKeySpace).stream()
```
maciejgrzybek: Good point! Added in 9e4573db5809031cf5506686025b8f7904c7ffed.

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraTableManager.java
```diff
@@ -0,0 +1,209 @@
...
+        List<String> columnNames = queryExecutor.getColumnNames(tableName.getSchema().get(), tableName.getInternalNameInDatabase());
+
+        try {
+            CassandraBatchLoader loader = new CassandraBatchLoader(queryExecutor.getSession(), tableName.getNameInDatabase(), columnNames);
+            for (List<List<Object>> batch : partitionBy(dataRows, BATCH_SIZE)) {
+                loader.load(batch);
+            }
+        }
+        catch (SQLException e) {
+            throw new RuntimeException(e);
```
maciejgrzybek: This is the same as it was done in `JdbcTableManager`. I simply copy&pasted it.
Though, `Throwables.propagate` seems better approach.
Fixed in .

tempto-core/src/main/java/com/teradata/tempto/internal/query/CassandraQueryExecutor.java
maciejgrzybek: Right, I forgot about that change.

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraBatchLoader.java
```diff
@@ -0,0 +1,67 @@
...
+    {
+        PreparedStatement statement = session.prepare(format("INSERT INTO %s (%s) VALUES(%s)",
+                tableName,
+                columnNames.stream().collect(joining(",")),
+                repeatPattern("?", ",", columnNames.size())));
+
+        BatchStatement batchStatement = new BatchStatement(BatchStatement.Type.UNLOGGED);
+
+        for (List<Object> row : batch) {
+            batchStatement.add(statement.bind(row.toArray()));
```
arhimondr: flush it per `N` row

tempto-core/src/main/java/com/teradata/tempto/fulfillment/table/jdbc/RelationalDataSource.java
maciejgrzybek: There's a difference between `HiveTableDefinition` and `RelationalTableDefinitions` because there are some Hadoop specific things which we want to express there, e.g. file type, compression. I tried to rewrite that but it was PITA because `HiveTableDefitnition` heavily depends on the fact that data source is of type `HiveDataSource`. And even if it's not a `HiveTableDefinition` depending on that directly it's exposed further through `getDataSource()` method and it's used in e.g. `HiveTableManager`. This dependency simply leaks further...

Anyway, it's way beyond the scope of this PR. For now Cassandra and JDBC databases uses exactly the same properties to express tables so I don't see the reason to not reuse that code. When we notice significant differences between `JDBCTableDefinition` and `CassandraTableDefinition` we can rename back `RelationalTableDefinition` to `JDBCTableDefinition` and introduce `CassandraTableDefinition`. For now I believe we should reuse code which would be identical anyway...

tempto-core/src/main/java/com/teradata/tempto/internal/query/QueryExecutorModuleProvider.java
```diff
@@ -54,7 +54,9 @@ protected void configure()
                 bind(JdbcConnectionsPool.class).toInstance(jdbcConnectionsPool);
                 Set<String> definedJdcbConnectionNames = jdbcConnectionsConfiguration.getDefinedJdcbConnectionNames();
                 for (String connectionName : definedJdcbConnectionNames) {
-                    bindDatabaseConnectionBeans(connectionName);
+                    if (!connectionName.equals("cassandra")) {
```
maciejgrzybek: I don't necessarily agree here.
Cassandra has nothing to do with JDBC so I don't think we should hardcode exception for it inside JDBC "internal" classes. It's hacked on the quite generic level.
I agree we should not be binding `QueryExecutor` if it's not exposed nor needed but I claim it's outside of scope of this PR.

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraTableManager.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    @Override
+    public TableInstance<RelationalTableDefinition> createImmutable(RelationalTableDefinition tableDefinition, TableHandle tableHandle)
+    {
+        TableName tableName = createImmutableTableName(tableHandle);
+        checkArgument(tableName.getSchema().isPresent(), "Cassandra does not like you when you don't have keyspace defined");
+
+        if (!queryExecutor.tableExists(tableName.getSchema().get(), tableName.getInternalNameInDatabase())) {
+            executeQueryIgnoreError(tableDefinition.getCreateTableDDL(tableName.getNameInDatabase()));
```
arhimondr: Why `ignoreError?`

arhimondr:
lgtm % comments

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraTableManager.java
maciejgrzybek: We don't support all data types in Cassandra.
`CassandraQueryExecutor.getJDBCType` throws an exception when there's no mapping for data type. We don't care about output of `executeQuery(SQL)` in DDL because we want to be able to create a table even with types we don't support.
Though `CassandraQueryExecutor` should throw that exception in case one wanted to read that output.

tempto-core/src/main/java/com/teradata/tempto/fulfillment/table/jdbc/RelationalDataSource.java
kokosing: ok, ok, ok ;)

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/TableName.java
kokosing: So maybe rename this method to `getSchemalessNameInDatabase()`?

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraTableManager.java
```diff
@@ -0,0 +1,228 @@
...
+
+    private static final int BATCH_SIZE = 10_000;
+
+    private final TableNameGenerator tableNameGenerator;
+    private final CassandraQueryExecutor queryExecutor;
+    private final String databaseName;
+    private final String defaultKeySpace;
+
+    @Inject
+    public CassandraTableManager(TableNameGenerator tableNameGenerator,
```
kokosing: each parameter in separate line

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraTableManager.java
```diff
@@ -0,0 +1,228 @@
...
+        this.queryExecutor = new CassandraQueryExecutor(configuration);
+        this.databaseName = requireNonNull(databaseName, "databaseName is null");
+        this.defaultKeySpace = configuration.getStringMandatory("databases.cassandra.default_schema");
+    }
+
+    @Override
+    public TableInstance<RelationalTableDefinition> createImmutable(RelationalTableDefinition tableDefinition, TableHandle tableHandle)
+    {
+        TableName tableName = createImmutableTableName(tableHandle);
+        checkArgument(tableName.getSchema().isPresent(), "Cassandra does not like you when you don't have keyspace defined");
```
kokosing: I would rephrase the message to "Cassandra requires table with schema". I don't think it is a personal Cassandra's feeling, especially to that particular user.

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraTableManager.java
kokosing: it sounds like a bad idea. If you ignore error from `CREATE TABLE`, how can you insert data two lines below.

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraTableManager.java
```diff
@@ -0,0 +1,228 @@
...
+            for (List<List<Object>> batch : partitionBy(dataRows, BATCH_SIZE)) {
+                loader.load(batch);
+            }
+        }
+        catch (SQLException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private static Iterable<List<List<Object>>> partitionBy(Iterator<List<Object>> dataRows, int partitionSize)
```
kokosing: extract some entities to limit the amount of copy-pasted code from `JdbcTableManger`

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraTableManager.java
```diff
@@ -216,12 +216,12 @@ protected TableName createImmutableTableName(TableHandle tableHandle)
...
     }
 
-    private void executeQueryIgnoreError(String sql)
+    private void executeQueryIgnoreTypeError(String sql)
     {
         try {
             queryExecutor.executeQuery(sql);
         }
-        catch (IllegalStateException e) {
+        catch (CassandraQueryExecutor.TypeNotSupportedException e) {
```
kokosing: Can you check the type before running the query? I prefer using `if` than `try-catch`

maciejgrzybek: This is a part of `QueryExecutor` logic. I don't thinkg `CassandraTableManager` should know what types `QueryExecutor` supports. Especially while the output of `executeQuery` is not used by `CassandraTableManager`.

I could return nothing from `CassandraQueryExecutor.executeQuery` as well but I wanted to keep it similar to other `QueryExecutors`, if anyone needed it outside of `CassandraTableManager`.

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraTableManager.java
maciejgrzybek: It's only two places... And it has to be rewritten heavily when we'll introduce a proper modularization.

maciejgrzybek:
AD https://github.com/prestodb/tempto/pull/180#discussion_r86135171: fixed in 60faa2d7d2e22919e605564d440eb6235f662b95.
Will rename commit while squashing.

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraTableManager.java
maciejgrzybek: Fixed in: 272bb42cae0fad880864d6db06ce544aeeda6e60.

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraTableManager.java
maciejgrzybek: e1dcc8adab48a9a6c55dae4b64bb9bc83ee8d1b2

maciejgrzybek:
@kokosing given how cumbersome it was to modularize Cassandra properly (decouple test initialization etc.), I abandoned that idea in favor of delivering testing capabilities faster.
We should plan for proper modularization before next module is introduced but we should not spend more time on it now given testing Cassandra is quite urgent issue for now.

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/TableName.java
maciejgrzybek: Done in 60faa2d7d2e22919e605564d440eb6235f662b95.

maciejgrzybek:
Commits squashed per @arhimondr request.

arhimondr:
lgtm

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/tpch/CassandraTpchDataSource.java
```diff
@@ -0,0 +1,87 @@
...
+    implements RelationalDataSource
+{
+    private static final Splitter SPLITTER = Splitter.on('|');
+
+    private final TpchTable table;
+    private final List<Integer> mapping;
+    private final List<JDBCType> columnTypes;
+    private final double scaleFactor;
+
+    public CassandraTpchDataSource(TpchTable table, List<Integer> mapping, List<JDBCType> columns, double scaleFactor)
```
cawallin: What does this look like from a user perspective? You pass in a list of integers that specify the order that you want the tpch data to come out? I think it would be clearer from a user perspective for them to specify only the primary key(s) and handle the sorting here.

maciejgrzybek: Unfortunately you cannot get that list. TpchEntity is an interface not provided such an information.
Each implementation of that interface is providing different set of getters and different implementation of `toLine` method but no info about columns.
I know it's not clean but cleaner approach requires un-generalization of data sources and table managers which is quite expensive and outside of the scope of this PR.

maciejgrzybek:
Merged, thanks!
Further requests and changes should be addressed in separate PRs.

@kokosing: let's address your concerns as a separate refactor.

PR #181, opened 2016-11-03
Cassandra: Create schema (keySpace) when it's missing

maciejgrzybek:


tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraTableManager.java
```diff
@@ -78,6 +80,9 @@ public CassandraTableManager(
         checkArgument(tableName.getSchema().isPresent(), "Cassandra requires table with schema");
```
cawallin: Should this give the default keyspace or is that handled elsewhere?

maciejgrzybek: Well, the default schema was meant for removing stale tables. But we can use it for the same purpose.

maciejgrzybek:
Merged, thanks!

PR #184, opened 2016-11-07
Add simple SELECT java based example test for Cassandra

maciejgrzybek:


tempto-examples/docker/docker-compose.yml
```diff
@@ -30,6 +30,13 @@ services:
       POSTGRES_USER: blah
       POSTGRES_PASSWORD: blah
 
+  cassandra:
```
cawallin: do you need to add cassandra to run_on_docker?

tempto-examples/src/main/java/com/teradata/tempto/examples/CassandraQueryTest.java
```diff
@@ -0,0 +1,77 @@
...
+            return immutableTable(TEST_TABLE_DEFINITION);
+        }
+    }
+
+    @Test(groups = "cassandra_query")
+    @Requires(CassandraQueryTest.ImmutableTestCassandraTable.class)
+    public void selectFromImmutableTable()
+    {
+        // Cassandra disregards the order of columns outside primary key
+        // Primary key comes first, in the order as it's defined, rest of columns are ordered alphabetically
```
cawallin: Maybe add another element in the primary key to show this all?

tempto-examples/docker/docker-compose.yml
maciejgrzybek: How differently am I suppose to run an example? Tempto-examples are executed as a part of CI.

tempto-examples/src/main/java/com/teradata/tempto/examples/CassandraQueryTest.java
maciejgrzybek: How about 4e2a301?

maciejgrzybek:
@cawallin: pushed 77e563f to clarify that better.
The order of input should be the same as order of columns in Cassandra, not in DDL.

tempto-examples/docker/docker-compose.yml
maciejgrzybek: Ouch, sorry. I didn't get your point here. No, we don't need to add it to the `run_on_docker.sh` script. It calls `${DOCKER_COMPOSE} up -d` which starts all dockers in the same compose.

maciejgrzybek:
Merged, thanks!

PR #185, opened 2016-11-07
Create schema if not exists for mutable tables in Cassandra

maciejgrzybek:
Similarly to immutable tables.


maciejgrzybek:
Merged, thanks!

PR #187, opened 2016-11-10
Add dedicated TableDefinition for Cassandra

maciejgrzybek:
This enables using Cassandra product tests in the same suite
along with other relational requirements (e.g. MySQL, PSQL etc.).

### DIFF
--- a/tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraTableDefinition.java
+++ b/tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraTableDefinition.java
@@ -1,0 +1,79 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraTableDefinition.java
```diff
@@ -0,0 +1,79 @@
...
+import com.teradata.tempto.fulfillment.table.TableHandle;
+import com.teradata.tempto.fulfillment.table.jdbc.RelationalDataSource;
+import com.teradata.tempto.fulfillment.table.jdbc.RelationalTableDefinition;
+
+import static com.teradata.tempto.fulfillment.table.TableHandle.tableHandle;
+
+public class CassandraTableDefinition
+    extends RelationalTableDefinition
+{
+    public CassandraTableDefinition(String name, String createTableDDLTemplate, RelationalDataSource dataSource)
```
cawallin: Unless you have a particular reason, it'd be better to use a static method instead of a public constructor to mirror how RelationalTableDefinition works.

maciejgrzybek:
Merged, thanks!

PR #196, opened 2016-12-07
Remove LIMIT for Batch Loader

sanjay990:
LIMIT doesn't exist for SQL Server, hence we need a different method to
get the column count for creating Batch Loader

### DIFF
--- a/tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/LoaderFactory.java
+++ b/tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/LoaderFactory.java
@@ -18,8 +18,11 @@ import com.teradata.tempto.query.QueryExecutor;
 import org.slf4j.Logger;
 
 import java.sql.JDBCType;
+import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 
 import static org.slf4j.LoggerFactory.getLogger;
 
@@ -27,10 +30,16 @@ class LoaderFactory
 {
     private static final Logger LOGGER = getLogger(LoaderFactory.class);
 
-    Loader create(QueryExecutor queryExecutor, String tableName)
+    Loader create(QueryExecutor queryExecutor, Optional<String> schema, String tableName)
             throws SQLException
     {
-        List<JDBCType> columnTypes = queryExecutor.executeQuery("SELECT * FROM " + tableName + " LIMIT 1").getColumnTypes();
+        schema = getSchema(queryExecutor);
+        ResultSet resultSet = queryExecutor.getConnection().getMetaData().getColumns(null, schema.orElse(""), tableName, null);
+
+        List<JDBCType> columnTypes = new ArrayList<>();
+        while (resultSet.next()) {
+            columnTypes.add(JDBCType.valueOf(resultSet.getInt("DATA_TYPE")));
+        }
 
         try {
             return new BatchLoader(queryExecutor, tableName, columnTypes.size());
@@ -40,4 +49,14 @@ class LoaderFactory
             return new InsertLoader(queryExecutor, tableName, columnTypes);
         }
     }
+
+    private Optional<String> getSchema (QueryExecutor queryExecutor){
+        try {
+            return Optional.ofNullable(queryExecutor.getConnection().getSchema());
+        }
+        catch (Exception e) {
+            return Optional.empty();
+        }
+}
+
 }
--- a/tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/LoaderFactory.java
+++ b/tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/LoaderFactory.java
@@ -18,11 +18,8 @@ import com.teradata.tempto.query.QueryExecutor;
 import org.slf4j.Logger;
 
 import java.sql.JDBCType;
-import java.sql.ResultSet;
 import java.sql.SQLException;
-import java.util.ArrayList;
 import java.util.List;
-import java.util.Optional;
 
 import static org.slf4j.LoggerFactory.getLogger;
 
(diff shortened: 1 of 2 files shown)


tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/LoaderFactory.java
```diff
@@ -18,19 +18,30 @@
...
 
 class LoaderFactory
 {
     private static final Logger LOGGER = getLogger(LoaderFactory.class);
 
-    Loader create(QueryExecutor queryExecutor, String tableName)
+    Loader create(QueryExecutor queryExecutor, Optional<String> schema, String tableName)
             throws SQLException
     {
-        List<JDBCType> columnTypes = queryExecutor.executeQuery("SELECT * FROM " + tableName + " LIMIT 1").getColumnTypes();
```
KBP-TDC: Btw. I think SQL Server has "TOP" instead:
SELECT TOP (1) * FROM tableName
Docs: https://msdn.microsoft.com/en-us/library/ms189463.aspx

sanjay990: We plan to have something which doesn't depend on the underlying database.

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/LoaderFactory.java
```diff
@@ -18,19 +18,30 @@
...
+        if (!schema.isPresent() && queryExecutor.getConnector().isPresent() && queryExecutor.getConnector().get().equals("sqlserver")) {
```
cawallin: you shouldn't assume that the sqlserver connector is named sqlserver -- why do you need this? and why is it only for sqlserver?

sanjay990: For SQL Server, I need to get the default schema, else getColumns won't work
getSchema was not working for Postgres - org.postgresql.jdbc4.Jdbc4Connection.getSchema() is not yet implemented

tempto-core/src/main/java/com/teradata/tempto/query/QueryExecutor.java
```diff
@@ -56,6 +57,8 @@ QueryResult executeQuery(String sql, QueryType queryType, QueryParam... params)
 
     Connection getConnection();
 
+    Optional<String> getConnector();
```
maciejgrzybek: What does `connector` mean in this context? It's unclear.

sanjay990: It means presto connector like mysql, postgres, sqlserver

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/LoaderFactory.java
sanjay990: yes, but it doesn't work for PostgreSQL - org.postgresql.jdbc4.Jdbc4Connection.getSchema() is not yet implemented

tempto-core/src/main/java/com/teradata/tempto/query/QueryExecutor.java
maciejgrzybek: Then `ConnectorName` name? Or rather `CatalogName` if it's named connector?

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/LoaderFactory.java
sanjay990: no, schema is only needed for SQL Server, its not required for Postgres, MySQL

tempto-core/src/main/java/com/teradata/tempto/query/QueryExecutor.java
sanjay990: `CatalogName` seems to be a good choice

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/LoaderFactory.java
kokosing: lets wrap`org.postgresql.jdbc4.Jdbc4Connection.getSchema()` so when it throws an exception `queryExecutor.getConnection().getSchema()` would return `Optional.empty()`. Then remove this `IF`.

tempto-core/src/main/java/com/teradata/tempto/query/QueryExecutor.java
kokosing: I don't like this method. Maybe it has some meaning in regards to presto connection but it cann't be mapped to anything in case of hive connection. I think that when you fix my other comment you will be able to remove this method.

sanjay990:
@maciejgrzybek @kokosing @cawallin : Removed the sqlserver check

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/LoaderFactory.java
```diff
@@ -33,9 +33,13 @@
     Loader create(QueryExecutor queryExecutor, Optional<String> schema, String tableName)
             throws SQLException
     {
-        if (!schema.isPresent() && queryExecutor.getConnector().isPresent() && queryExecutor.getConnector().get().equals("sqlserver")) {
+        try {
             schema = Optional.ofNullable(queryExecutor.getConnection().getSchema());
         }
+        catch (Exception e) {
+            schema = Optional.empty();
+        }
```
kokosing: extract this `try-catch` block as separate method

kokosing:
Looks good, squash the commits and merge

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/jdbc/LoaderFactory.java
```diff
@@ -30,7 +30,7 @@
     Loader create(QueryExecutor queryExecutor, String tableName)
             throws SQLException
     {
-        List<JDBCType> columnTypes = queryExecutor.executeQuery("SELECT * FROM " + tableName + " LIMIT 1").getColumnTypes();
+        List<JDBCType> columnTypes = queryExecutor.executeQuery("SELECT * FROM " + tableName + " WHERE 1=2").getColumnTypes();
```
amrutagokhale: Is `getColumnCount()` not working? I feel that this method is more appropriate (if it works), since it conveys directly what the code is doing.

sanjay990: But we need the column types for insert loader

PR #197, opened 2016-12-14
Change default TPCH scale factor to 0.01

ebd2:


maciejgrzybek:
You need a preparing-next-development-cycle commit after release commit to not make automation unhappy.

maciejgrzybek:
Looks good % adding next dev cycle commit (ver = 1.23-SNAPSHOT).

PR #203, opened 2017-01-23
Update to TestNG 6.10

electrum:


tempto-runner/src/main/java/com/teradata/tempto/runner/TemptoRunner.java
```diff
@@ -133,7 +133,7 @@ private XmlSuite getXmlSuite()
         XmlClass conventionBasedTestsClass = new XmlClass("com.teradata.tempto.internal.convention.ConventionBasedTestFactory");
         List<XmlClass> classes = newArrayList(conventionBasedTestsClass);
         test.setClasses(classes);
-        test.setParallel(XmlSuite.PARALLEL_METHODS);
+        test.setParallel(XmlSuite.ParallelMode.METHODS);
```
electrum: We could add reflection here to call the old method if we want to be compatible with both versions. It's unfortunate they didn't keep the old method around for a least one release :(

electrum: This change seems to be introduced in `6.9.7`

electrum:
I had to add an explicit dependency on junit due to `SimpleQueryTest`. Is it supposed to be using the junit `@Before` / `@After` annotations, or it should it be using the ones from TestNG?

cawallin:
``SimpleQueryTest`` should be using the TestNG ``@Before`` and ``@After``, but it seems like there are some other tests that are junit tests (ReflectionInjectorHelperTest.groovy, etc). If you want to clean those up, go ahead, but you don't need to.

maciejgrzybek:
Merged with TODO note about jUnit dependencies removal. Will publish snapshot version momentarily.

PR #206, opened 2017-01-31
Add content-type header to WebHdfs requests

petroav:
- This change is needed because MapR uses HttpFS to
  interact with WebHDFS and that requires the content-type
  header to be included.


petroav:
How would you like me to test this? I ran the non-kerberized CDH product tests locally and they passed. I could release a 1.26-SNAPSHOT version with my branch and then run Presto checks using that version?

kokosing:
tempto examples passes so it is ok to me.

petroav:
Thanks for the review, merged!

PR #207, opened 2017-02-08
Decrease cassandra batch size

arhimondr:


tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/cassandra/CassandraTableManager.java
```diff
@@ -67,6 +66,7 @@ public CassandraTableManager(
         this.databaseName = requireNonNull(databaseName, "databaseName is null");
         this.defaultKeySpace = configuration.getStringMandatory("databases." + databaseName + ".default_schema");
         this.skipCreateSchema = configuration.getBoolean("databases." + databaseName + ".skip_create_schema").orElse(false);
+        this.insertBatchRowsCount = configuration.getInt("databases." + databaseName + ".insert_batch_rows_count").orElse(10);
```
maciejgrzybek: Why 10?
If you're afraid of exceeding limit, make it "1" - the safest possible option.
10 still does not help much with performance.

PR #213, opened 2017-03-28
Use Lazy<CassandraQueryExecutor> in CassandraTableManager

kokosing:
Use Lazy<CassandraQueryExecutor> in CassandraTableManager

Thanks to that when CassandraTableManager is not effectively used, then
it does not require to instantiate the CassandraQueryExecutor.

### DIFF
--- a/tempto-core/src/main/java/com/teradata/tempto/util/Lazy.java
+++ b/tempto-core/src/main/java/com/teradata/tempto/util/Lazy.java
@@ -14,10 +14,14 @@
 
 package com.teradata.tempto.util;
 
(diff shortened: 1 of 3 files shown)


tempto-core/src/main/java/com/teradata/tempto/util/Lazy.java
```diff
@@ -37,4 +39,9 @@ public synchronized T get()
         }
         return instance;
     }
+
+    public Optional<T> lazyGet()
```
findepi: synchronized -- this class is thread-safe and guaranteed happens-before between producing value and consuming it

PR #214, opened 2017-03-28
Make mutable table suffix shorter

kokosing:
Make mutable table suffix shorter

### DIFF
--- a/tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/TableNameGenerator.java
+++ b/tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/TableNameGenerator.java
@@ -14,7 +14,7 @@
 
 package com.teradata.tempto.internal.fulfillment.table;
 
-import java.util.UUID;
+import static org.apache.commons.lang3.RandomStringUtils.random;
 
 public class TableNameGenerator
 {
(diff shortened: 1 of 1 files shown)


tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/TableNameGenerator.java
```diff
@@ -31,8 +34,8 @@ public boolean isMutableTableName(String tableNameInDatabase)
         return tableNameInDatabase.startsWith(MUTABLE_TABLE_NAME_PREFIX);
     }
 
-    private String randomUUID()
+    private String randomSuffix()
```
findepi: `static`

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/TableNameGenerator.java
```diff
@@ -31,8 +34,8 @@ public boolean isMutableTableName(String tableNameInDatabase)
...
     {
-        return UUID.randomUUID().toString();
+        return Integer.toString(abs(random.nextInt()), Character.MAX_RADIX);
```
findepi: Still can be (unlikely) negative.

Instead of hand-crafting, consider using  `RandomUtils.randomAlphanumeric(..)` from apache commons ?

PR #220, opened 2017-06-09
Implement AWS S3 client for Tempto

smdorval:


tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,384 @@
...
+            @Named(CONF_S3_BUCKET_NAME) String bucketName)
+    {
+        // Check null values and set object variables
+        checkNotNull(keyId, "aws api access key id is null");
+        checkNotNull(secret, "aws api secret access key is null");
+        this.bucketName = checkNotNull(bucketName, "s3 bucket name is null");
+
+        // Initialize AWS SDK credentials
+        BasicAWSCredentials credentials = null;
+        try {
```
losipiuk: extract `BasicAWSCredential buildAWSCredentials(keyId, secret)` to hide exception handling inside it

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,384 @@
...
+            throw new AmazonClientException("Cannot initialize credentials", e);
+        }
+
+        // Create S3 client
+        client = AmazonS3ClientBuilder.standard()
+                .withCredentials(new AWSStaticCredentialsProvider(credentials))
+                .build();
+    }
+
+    /**
```
losipiuk: sugg: drop javadoc

This applies to all javadocs here. Try to capture the actual responsibility of the method in its name.

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,384 @@
...
+     * @param path              path to get contents
+     * @param includeFolders    true/false to include folders or not
+     * @return
+     */
+    public ArrayList<String> getFolderContents(String path, boolean includeFolders)
+    {
+        ArrayList<String> objects = new ArrayList<String>();
+
+        // If no forward slash at end of path, add one
+        if (!path.substring(path.length() - 1).equals("/")) {
```
losipiuk: will fail for empty path.

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,384 @@
...
+            path += "/";
+        }
+
+        for (String object : getAllObjects(path)) {
+            // Remove input path from object name and determine the number of forward slashes left
+            String objectName = object.replace(path, "");
```
losipiuk: Adn what if `path` appears in `object` two times.
E.g. input path path is `/foo` and
we have `/foo/foo/foo` on S3.

Generally this logic is very hard to follow.
I suggest that you rewrite logic here and operate on path segments.
Use guava's `Splitter.on('/')` to split input path and results of `getAllObjects` to cut paths int parts.
Then for each entity coming from `getAllObjects` take N-th element (where N is number of segments in input path) and collapse everything to `Set`.

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,384 @@
...
+    }
+
+    /**
+     * Method to get folder contents (including folders)
+     *
+     * @param path              path to get contents
+     * @param includeFolders    true/false to include folders or not
+     * @return
+     */
+    public ArrayList<String> getFolderContents(String path, boolean includeFolders)
```
losipiuk: Return `List` not `ArrayList`.

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,384 @@
...
+                .build();
+    }
+
+    /**
+     * Method to get folder contents
+     *
+     * @param path  path to get contents
+     * @return      list of object names
+     */
+    public ArrayList<String> getFolderContents(String path)
```
losipiuk: `listFiles`

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,384 @@
...
+
+    /**
+     * Method to create folder in S3
+     *
+     * @param path  folder path to create
+     */
+    public void createFolder(String path)
+    {
+        // If no forward slash at end of path, add one
+        if (!path.substring(path.length() - 1).equals("/")) {
```
losipiuk: Extract `ensureTrailingSlashPresent`

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,384 @@
...
+
+    /**
+     * Method to get folder contents (including folders)
+     *
+     * @param path              path to get contents
+     * @param includeFolders    true/false to include folders or not
+     * @return
+     */
+    public ArrayList<String> getFolderContents(String path, boolean includeFolders)
+    {
```
losipiuk: the `includeFolders` flag is not used at all

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,384 @@
...
+            path += "/";
+        }
+
+        // If folder already exists, throw exception
+        if (exists(path)) {
+            throw new RuntimeException("Folder already exists - could not create folder: " + path);
+        }
+
+        // Create meta-data for your folder and set content-length to 0
+        ObjectMetadata metadata = new ObjectMetadata();
```
losipiuk: How an object is marked as a folder?

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,384 @@
...
+     */
+    public void deleteFolder(String path, boolean recursive)
+    {
+        // If no forward slash at end of path, add one
+        if (!path.substring(path.length() - 1).equals("/")) {
+            path += "/";
+        }
+
+        // Get all objects in this folder
+        ArrayList<String> objects = getAllObjects(path);
```
losipiuk: Use `List` for variable type.

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,384 @@
...
+        if (!isFolder) {
+            deleteObject(path);
+        } else {
+            throw new RuntimeException("Folder was passed in to deleteFile: " + path);
+        }
+
+        logger.debug("Deleted file {} from S3", path);
+    }
+
+    /**
```
losipiuk: drop this

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,384 @@
...
+     * PRIVATE METHODS
+     */
+
+    /**
+     * Method to get all object names in given path
+     *
+     * @param path  path to search
+     * @return      list of object names
+     */
+    public ArrayList<String> getObjects(String path)
```
losipiuk: why "private" methods are public?

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
losipiuk: Why `getAllObjects`? Do we need to do recursive search here?

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,384 @@
...
+        logger.debug("Deleted S3 folder {}", path);
+    }
+
+    /**
+     * Method to save file from direct local file
+     *
+     * @param path      path on S3 to save
+     * @param filePath  path of input file to save
+     */
+    public void saveFile(String path, String filePath)
```
losipiuk: I do not like the idea of accessing local filesystem from S3 specific client.
The source of data written to S3 should not be concern of this class.

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,384 @@
...
+     * Method to save file from input stream
+     *
+     * @param path          path on S3 to save
+     * @param stream        input stream
+     * @param byteLength    length of data (in bytes)
+     */
+    public void saveFile(String path, InputStream stream, long byteLength)
+    {
+        ObjectMetadata metadata = new ObjectMetadata();
+        metadata.setContentLength(byteLength);
```
losipiuk: Do we need to know size of content upfront?
How do we validate that `stream` actually provider proper amount of data. Does wrapped `S3` client does that?

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,384 @@
...
+        }
+
+        // Get all objects in this folder
+        ArrayList<String> objects = getAllObjects(path);
+
+        if (!recursive && objects.size() > 0) {
+            throw new RuntimeException("Folder is not empty - could not remove folder: " + path);
+        } else if (objects.size() > 0) {
+            for(String objectName : objects) {
+                deleteObject(objectName);
```
losipiuk: is the order of deletions here proper. I assume that `deleteObject` is not recursive, as otherwise we would not need to do this iteration here at all. So the question is are paths on `objects` list ordered in such way that we will delete children first and parent later?

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,384 @@
...
+
+    /**
+     * Method to deletes object
+     *
+     * @param path  path of object to delete
+     * @throws RuntimeException
+     */
+    private void deleteObject(String path)
+    {
+        boolean folder = path.substring(path.length() - 1).equals("/");
```
losipiuk: determining if path points to folder or file just to have different exception is not worth an effort. Drop that.
And make exception message `Object does not exist`.

The other question is do we actually want to fail the `deleteObject` method if we remove inexistent object. This can make use of such method in cleanup procedures problematic.

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,384 @@
...
+        return objects;
+    }
+
+    /**
+     * Method to get all object names in given path (recursively)
+     *
+     * @param path  path to search
+     * @return      list of object names
+     */
+    public ArrayList<String> getAllObjects(String path)
```
losipiuk: List

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,384 @@
...
+    private static final Logger logger = getLogger(S3Client.class);
+
+    public static final String CONF_S3_AUTH_KEY_ID = "s3.auth.keyid";
+    public static final String CONF_S3_AUTH_SECRET = "s3.auth.secret";
+    public static final String CONF_S3_BUCKET_NAME = "s3.bucketname";
+
+    private final String bucketName;
+    private final AmazonS3 s3client;
+
+    /**
```
smdorval: ok, makes sense - i will remove all of the javadoc stuff

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
smdorval: what do you mean by this?

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
smdorval: the flag is there to differentiate between the 2 functions of the same name - one that lists only files and the other that lists files and folders - would it be better to just have 2 different names for the functions like listFiles() and listAll()?

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
smdorval: i will put a check in there for empty path

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
losipiuk: Extract:

```java
private BasicAWSCredentials buildAwsCredentials(String keyId, String secret) {
    try {
        credentials = new BasicAWSCredentials(keyId, secret);
    } catch (Exception e) {
        throw new AmazonClientException("Cannot initialize credentials", e);
    }
}
```

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
losipiuk: I don't get it. The javadoc (and that is what you would guess based on parameter name) that depending if `includeFolders` is set to `true` or `false` the method behaviour is different.
But currently it is not as this parameter value is not read anywhere.

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
smdorval: got it - not sure what i was thinking

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
smdorval: The way S3 handles folders is that it is an object with zero length and a forward slash at the end.

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
smdorval: In S3 each item is an object with the name given.  For example, test/test.txt is one object and test/ is another object so it does not matter the order that you delete them.  If you delete test/ before test/test.txt, test/test.txt will still be there.  Make sense?

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
smdorval: This is removed.

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
smdorval: Unfortunately the S3 client does not calculate the proper amount of data and it needs to be provided.  In my tests I just read the size of the stream before I sent it to the new client.

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
smdorval: Dropped

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
smdorval: Fixed

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
smdorval: Make sense.  I removed the check for a file or folder and just delete it no matter what and we no longer fail if it tries to delete a non-existent object to help with the cleanup procedures.

smdorval:
@losipiuk you mentioned about some test code to test the interface.  I have some test code that I used locally on my computer.  How should that be included in the project?  Or should it not?

losipiuk:
>@losipiuk you mentioned about some test code to test the interface. I have some test code that I used locally on my computer. How should that be included in the project? Or should it not?

I meant some code in tempto which uses the `S3Client`. For table provisioning. Without that it is kinda dead code.

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,263 @@
...
+            S3Object object = client.getObject(new GetObjectRequest(bucketName, path));
+            IOUtils.copy(object.getObjectContent(), outputStream);
+        } catch (IOException exception) {
+            throw new RuntimeException("Could not read file " + path + " from S3", exception);
+        }
+
+        logger.debug("Loaded file {} from S3", path);
+    }
+
+    public String loadFile(String path)
```
losipiuk: Returning file contents (potentially binary) as `String` is bad idea. Use `ByteBuffer` or `byte[]` if method returning `OutputStream` is not enough. 
You may name this method `readFileEagerly` or sth like that to avoid naming clash with method above.

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,263 @@
...
+    {
+        ByteArrayOutputStream output = new ByteArrayOutputStream();
+        loadFile(path, output);
+        return new String(output.toByteArray(), defaultCharset());
+    }
+
+    public void deleteFile(String path)
+    {
+        boolean isFolder = path.substring(path.length() - 1).equals("/");
```
losipiuk: `path.endsWith('/')`

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,263 @@
...
+
+        // If not a folder, delete file; otherwise throw exception
+        if (!isFolder) {
+            deleteObject(path);
+        } else {
+            throw new RuntimeException("Folder was passed in to deleteFile: " + path);
+        }
+    }
+
+    /**
```
losipiuk: I do not think grouping public and private methods in blocks is best file organization. Typically the code is easiest to read if method appears  first just below the method which uses it for the first time. You can have exceptions from that for sure but it is good rule of thumb.

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,263 @@
...
+        try {
+            return new BasicAWSCredentials(keyId, secret);
+        } catch (Exception e) {
+            throw new AmazonClientException("Cannot initialize credentials", e);
+        }
+    }
+
+    private String ensureTrailingSlashPresent(String path)
+    {
+        if (path == "") {
```
losipiuk: replace content withj
```
if (!path.endsWith('/')) {
    path = path + '/';
}
return path;
```

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,263 @@
...
+        } else if (!path.substring(path.length() - 1).equals("/")) {
+            path += "/";
+        }
+
+        return path;
+    }
+
+    private String ensureTrailingSlashRemoved(String path)
+    {
+        if (path != "" && path.substring(path.length() - 1).equals("/")) {
```
losipiuk:
```
if (path.endsWith('/')) {
   path = path.substring(path.length() -1);
}
return path;
```

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,263 @@
...
+        ObjectMetadata metadata = new ObjectMetadata();
+        metadata.setContentLength(0);
+
+        InputStream emptyContent = new ByteArrayInputStream(new byte[0]);
+        PutObjectRequest putObjectRequest = new PutObjectRequest(bucketName,
+                ensureTrailingSlashPresent(path), emptyContent, metadata);
+
+        // Send request to S3 to create folder
+        client.putObject(putObjectRequest);
+
```
losipiuk: remove empty line

smdorval:
That is coming next.  I need to implement the TableManager and TableDefinition for Hive on S3.

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    private List<String> getObjects(String path)
+    {
```
losipiuk: what about
```java
    ListObjectsV2Result listing = client.listObjectsV2(new ListObjectsV2Request()
            .withBucketName(bucketName)
            .withPrefix(ensureTrailingSlashPresent(path))
            .withDelimiter("/"));
   
    return listing.getObjectSummaries().stream()
              .map(S3ObjectSummary::getKey)
              .filter(key -> !key.endsWith('/'))
              .collect(Collectors.toList());
```

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,263 @@
...
+
+    private List<String> getAllObjects(String path)
+    {
+        List<String> objects = new ArrayList<String>();
+
+        ListObjectsV2Result listing = client.listObjectsV2(new ListObjectsV2Request()
+                .withBucketName(bucketName)
+                .withPrefix(ensureTrailingSlashPresent(path)));
+
+        for (S3ObjectSummary objectSummary : listing.getObjectSummaries()) {
```
losipiuk: consider approach using stream like suggested above

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,232 @@
...
+    {
+        ObjectMetadata metadata = new ObjectMetadata();
+        metadata.setContentLength(byteLength);
+
+        client.putObject(new PutObjectRequest(bucketName, path, stream, metadata));
+
+        logger.debug("Saved file {} to S3", path);
+    }
+
+    public OutputStream readFile(String path)
```
losipiuk: I misguided you here. It should return `InputStream`, not `OutputStream` as you are going to read the file's data from it in code which calls `readFile`.

So that is why I suggested to just return `object.getObjectContent()`.

If you want to read the file's content within method to some array as you do it here
you should return not `output` from the method. But `new ByteArrayInputStream(output.toByteArray[])`.
I am not convinced that it is needed here though.

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    public boolean exists(String path)
+    {
+        return client.doesObjectExist(bucketName, path);
+    }
+
+    public List<String> listFiles(String path)
+    {
+        // Get list of all files located in this path
```
losipiuk: exactly the same statement in `listFolders` is commented with `// get list of all folders located in this path`.
This is misleading.

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
```diff
@@ -0,0 +1,232 @@
...
+
+    private String ensureTrailingSlashPresent(String path)
+    {
+        if (!path.endsWith(PATH_SEPARATOR)) {
+            path = path + PATH_SEPARATOR;
+        }
+        return path;
+    }
+
+    public List<String> listFolders(String path)
```
losipiuk: which part of this code does filtering out folders out of list of all objects?

Probably
```
+        // Return all objects
 +        return listing.getCommonPrefixes().stream()
 +                .collect(Collectors.toList());
```

But I do not understand what it does. Could you explain?

losipiuk: Also comment `// Return all objects` seems wrong.

losipiuk:
As for:

> @losipiuk That is coming next. I need to implement the TableManager and TableDefinition for Hive on S3.

Without that I am not convinced that interface of `S3Client` does the job. As we finish round of review of this one please squash the commits related to `S3Client` into one and add implementation of `TableManager` into same PR.
It may happen that during implementation of that you will need to change details of `S3Client` interface. And by working on client and user of client in single PR we can push it as single coherent entity, without unnatural fixup commits.

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
smdorval: I will change the comment to get list of all objects located in this path for both because that is what it is doing

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
smdorval: The getCommonPrefixes() method of the ListObjectsV2Result object returns all folders within the passed in prefix.  So for example if you had a directory structure like,

foo/bar/baz
foo/bar/bash
foo/boo

It will only return "foo/bar" since it is the only folder (common prefix) listed.

Modified the comment to say "Return all folders located in this path"

tempto-core/src/main/java/com/teradata/tempto/s3/S3Client.java
smdorval: Modified this to return the InputStream instead of the OutputStream.  I actually thought the same thing that it should return an InputStream but forgot to go back and check on this.

PR #222, opened 2017-08-03
Fix time zone support in Tempto

fiedukow:


losipiuk:
Be more descriptive in commit message. And split the commit into two.

tempto-core/src/main/java/com/teradata/tempto/util/DateTimeUtils.java
```diff
@@ -44,5 +45,17 @@ public static Timestamp parseTimestampInUTC(String value)
...
     }
+
+    /**
+     * Parse the input date time string (pattern: yyyy-M-d H:m:s.SSS) and interpret it in local time zone.
+     *
+     * @param value textual date time input
+     * @param timeZone local time zone
+     * @return a <code>java.sql.Timestamp</code> object representing the input date time.
+     */
+    public static Timestamp parseTimestampInLocalTimeZone(String value, DateTimeZone timeZone)
```
losipiuk: Is that used in tempto. If not is should not be defined here.

fiedukow:
done

PR #229, opened 2017-08-16
Pull docker images quietly and in parallel

kokosing:
Pull docker images quietly and in parallel


kokosing:
buuu travis is using too old docker-compose

ArturGajowy:
You can use a custom version of docker compose if needed :)

PR #233, opened 2017-08-22
Fix stats for tpch.lineitem and tpch.partsupp

kokosing:
Fix stats for tpch.lineitem and tpch.partsupp


kokosing:
This covers the issues I have found when running current tempto with presto.

PR #235, opened 2017-08-29
Refactor and changes for support of multiple file systems

smdorval:
Refactoring of tempto code and additions to support multiple file systems for Hive data storage


tempto-core/src/main/java/com/teradata/tempto/hadoop/FileSystemClient.java
```diff
@@ -45,6 +45,8 @@ InputStream getInputStream()
 
     void createDirectory(String path);
 
+    void deleteDirectory(String path);
```
kokosing: what is a difference between these two methods? If first is able to delete directories and second only files, then second should fail when getting a directory, so you cannot implement the first one with the second.

As I understand previous semantic was to delete always (files and directories), was there any issue with such approach? Can we leave it as it was?

tempto-examples/src/main/resources/tempto-configuration.yaml
```diff
@@ -77,6 +77,8 @@ databases:
 
 tests:
   fs:
+    type: hdfs
+    prefix: ""
```
kokosing: do I need to specify this if it is empty?

tempto-examples/src/main/resources/tempto-configuration.yaml
```diff
@@ -77,6 +77,8 @@ databases:
 
 tests:
   fs:
+    type: hdfs
```
kokosing: I do not see any usages of this.

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/hive/HiveTableManager.java
```diff
@@ -60,6 +61,7 @@ public HiveTableManager(
             QueryExecutor queryExecutor,
             FileSystemDataSourceWriter fsDataSourceWriter,
             TableNameGenerator tableNameGenerator,
+            @Named("tests.fs.prefix") String testDataBaseFSPrefix,
```
kokosing: Can you please explain me what it is? It looks we need to update documentation to cover this configuration option.

tempto-core/src/main/java/com/teradata/tempto/hadoop/FileSystemClient.java
smdorval: At first I thought that this was not possible because of the way directories are represented in S3 and Azure but now I think this is possible.  I will investigate.

tempto-examples/src/main/resources/tempto-configuration.yaml
smdorval: The usage of this will be visible when I add the S3 client changes on top of these commits.  Should I add that in the same PR?  I have the changes already and just need to finish some testing.

tempto-examples/src/main/resources/tempto-configuration.yaml
smdorval: Good catch.  This can probably be written that if it is not defined then it will default to "".

tempto-core/src/main/java/com/teradata/tempto/internal/fulfillment/table/hive/HiveTableManager.java
smdorval: I will nee to update the documentation for this.  The prefix is what is needed for Hive for location storage.  For example, the prefix for S3 would be s3n://[bucketname] ..

kokosing:
closing this as it looks it is no longer needed

PR #236, opened 2017-12-12
Display current configuration before running tests

kokosing:
Display current configuration before running tests


tempto-core/src/main/java/com/teradata/tempto/internal/initialization/TestInitializationListener.java
```diff
@@ -172,6 +172,8 @@ public TestInitializationListener(
     @Override
     public void onStart(ITestContext context)
     {
+        displayConfigurationToUser();
```
losipiuk: nit: maybe just `logConfiguration`

PR #246, opened 2017-12-21
Remove unused QueryType

kokosing:
Remove unused QueryType

It was used to decide when to use execute or executeUpdate on JDBC
Statement class. It is no longer needed.


tempto-core/src/main/java/com/teradata/tempto/internal/convention/SqlQueryDescriptor.java
```diff
@@ -107,9 +104,4 @@ public String getDatabaseName()
     {
         return getPropertyValuesSet(REQUIRES_HEADER_PROPERTY);
     }
-
-    public Optional<QueryType> getQueryType()
```
findepi: delete also `QueryType` itself

findepi:
@kokosing see travis failures

PR #247, opened 2018-01-05
Exception messages

findepi:


tempto-core/src/main/java/com/teradata/tempto/assertions/QueryAssert.java
```diff
@@ -415,6 +415,25 @@ public QueryExecutionAssert failsWithMessage(String expectedErrorMessage)
 
             return this;
         }
+
+        public QueryExecutionAssert failsWithMessageMathing(String expectedErrorMessagePattern)
```
kokosing: typo s/Mathing/Matching

tempto-core/src/main/java/com/teradata/tempto/assertions/QueryAssert.java
```diff
@@ -415,6 +415,25 @@ public QueryExecutionAssert failsWithMessage(String expectedErrorMessage)
...
+            String exceptionMessage = executionException.getMessage();
+            LOGGER.debug("Query failed as expected, with message: {}", exceptionMessage);
+            if (!nullToEmpty(exceptionMessage).matches(expectedErrorMessagePattern)) {
+                throw new AssertionError(format(
+                        "Query failed with unexpected error message: '%s' \n Expected error message to match '%s'",
+                        exceptionMessage,
+                        expectedErrorMessagePattern));
+            }
+
+            return this;
```
kokosing: this method is copy-paste of method above, maybe you could extract common code

tempto-core/src/main/java/com/teradata/tempto/assertions/QueryAssert.java
findepi: good catch!

tempto-core/src/main/java/com/teradata/tempto/assertions/QueryAssert.java
findepi: good idea, done

PR #248, opened 2018-01-12
Do not log configuration binding

kokosing:
Do not log configuration binding

Configuration is logged during the tempto startup once. There is no need
to log as DEBUG all the configuration entries per each test.

### DIFF
--- a/tempto-core/src/main/java/com/teradata/tempto/internal/configuration/TestConfigurationModuleProvider.java
+++ b/tempto-core/src/main/java/com/teradata/tempto/internal/configuration/TestConfigurationModuleProvider.java
@@ -25,7 +25,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.List;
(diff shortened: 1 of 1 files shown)


tempto-core/src/main/java/com/teradata/tempto/internal/configuration/TestConfigurationModuleProvider.java
```diff
@@ -50,17 +49,10 @@ protected void configure()
...
-                    if (value.isPresent()) {
-                        @SuppressWarnings("unchecked") Key bindingKey = getBindingKey(value.get(), key);
-
-                        LOGGER.debug("Binding {} key: {} -> {}", bindingKey, key, value.get());
-
-                        bind(bindingKey)
-                                .toInstance(value.get());
-                    }
-                }
+                configuration.listKeys().forEach(key ->
```
findepi: I think I'd prefer `{..}` after `key ->`.

PR #249, opened 2018-01-12
Formatting

kokosing:


kokosing:
> Is this just automated formatting with an IDE? any manual changes?

All are automated. There were some which I rejected (not included in this PR).

PR #252, opened 2018-01-24
Scan whole classpath in ReflectionHelper

losipiuk:
Previously list of jars which are scanned for classes (mostly for sake
of auto binding) was limited to io.*. This limits users of tempto and
actualy prevented tempto from working in Presto after the prefix was
chaned from com.* to io.*.

It seems like scanning whole classpath, though not fast, completes in
reasonable amount of time. Around 10s on macbook pro.

### DIFF
--- a/tempto-core/src/main/java/io/prestodb/tempto/internal/ReflectionHelper.java
+++ b/tempto-core/src/main/java/io/prestodb/tempto/internal/ReflectionHelper.java
@@ -29,12 +29,10 @@ import java.util.Set;
 import static java.util.Collections.unmodifiableSet;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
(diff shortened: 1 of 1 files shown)


tempto-core/src/main/java/io/prestodb/tempto/internal/ReflectionHelper.java
```diff
@@ -29,20 +29,18 @@
 import static java.util.Collections.unmodifiableSet;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
-import static org.reflections.util.ClasspathHelper.forPackage;
+import static org.reflections.util.ClasspathHelper.forJavaClassPath;
```
findepi: cmt msg:
`s/chaned /changed /`

PR #253, opened 2018-01-24
Remove cassandra related hacks + some cleanups

losipiuk:


build.gradle
```diff
@@ -27,7 +27,7 @@ plugins.withType(JavaPlugin) {
 
 ext.versions = [
         assertj      : '2.0.0',
-        guava        : '18.0',
+        guava        : '21.0',
```
kokosing: no changes? hm...

tempto-core/src/main/java/io/prestodb/tempto/internal/query/JdbcConnectionsConfiguration.java
```diff
@@ -26,7 +26,6 @@
 
 public class JdbcConnectionsConfiguration
 {
-
```
kokosing: can you please somehow remove all such entries?

build.gradle
losipiuk: Apparently. I updated as I wanted to have `toImmutableSet` method :)

PR #254, opened 2018-01-26
Allow fine control over injecting table statistics into Hive tables

findepi:
Having global switches `inject_stats_for_mutable_tables` and
`inject_stats_for_immutable_tables` is good, but not flexible enough.

### DIFF
--- a/tempto-core/src/main/java/io/prestodb/tempto/fulfillment/table/hive/HiveTableDefinition.java
+++ b/tempto-core/src/main/java/io/prestodb/tempto/fulfillment/table/hive/HiveTableDefinition.java
@@ -60,7 +60,7 @@ public class HiveTableDefinition
         return dataSource.get();
     }
 
-    public List<PartitionDefinition> getPartitionDefinitons()
+    public List<PartitionDefinition> getPartitionDefinitions()
     {
         checkState(isPartitioned(), "not supported for not partitioned table");
         return partitionDefinitions.get();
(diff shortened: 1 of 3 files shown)


tempto-core/src/main/java/io/prestodb/tempto/fulfillment/table/hive/HiveTableDefinition.java
```diff
@@ -46,7 +46,7 @@
     private HiveTableDefinition(TableHandle handle, String createTableDDLTemplate, Optional<HiveDataSource> dataSource, Optional<List<PartitionDefinition>> partitionDefinitions)
     {
         super(handle);
-        checkArgument(dataSource.isPresent() ^ partitionDefinitions.isPresent(), "either dataSource or partitionDefinitions must be set");
+        checkArgument(dataSource.isPresent() != partitionDefinitions.isPresent(), "either dataSource or partitionDefinitions must be set (but not both)");
```
kokosing: what is the difference?

findepi: Semantically, none.

kokosing: So what is the motivation, and why does the commit begin with `Fix`?

PR #262, opened 2018-04-05
Enhancements around convention tables

kokosing:


tempto-core/src/main/java/io/prestodb/tempto/internal/convention/ProcessUtils.java
```diff
@@ -14,19 +14,26 @@
...
+import java.util.function.Consumer;
 
 import static com.google.common.base.Preconditions.checkState;
 
 public final class ProcessUtils
 {
     private static final int SUCCESS_EXIT_CODE = 0;
 
-    public static void execute(String... cmdarray)
+    public static void execute(String ... cmdarray)
```
findepi: unnecessary change

tempto-core/src/main/java/io/prestodb/tempto/internal/convention/ProcessUtils.java
```diff
@@ -14,19 +14,26 @@
...
+    {
+        execute(process -> {}, cmdarray);
+    }
+
+    public static void execute(Consumer<Process> processConsumer, String ... cmdarray)
```
findepi: fmt

tempto-core/src/main/java/io/prestodb/tempto/internal/convention/tabledefinitions/ConventionTableDefinitionDescriptor.java
```diff
@@ -95,6 +106,16 @@ public ConventionTableDefinitionDescriptor(Path ddlFile)
         }
     }
 
+    private long copy(Path dataFile, Process process)
```
findepi: `s/copy/feed/`

tempto-core/src/main/java/io/prestodb/tempto/internal/convention/tabledefinitions/ConventionTableDefinitionDescriptor.java
```diff
@@ -83,6 +87,13 @@ public ConventionTableDefinitionDescriptor(Path ddlFile)
         this.ddlFile = ddlFile;
 
         Path dataFile = changeExtension(ddlFile, "data");
+        Path dataGeneratorFile = changeExtension(ddlFile, "data-generator");
+        if (exists(dataGeneratorFile)) {
+            checkArgument(!exists(dataFile), "Expected no data file when data-generator file is present");
+            makeExecutable(dataGeneratorFile);
```
findepi: ouch. This should be read-only.

kokosing: It is a temporary copy of read only files

tempto-core/src/main/java/io/prestodb/tempto/internal/convention/tabledefinitions/ConventionTableDefinitionDescriptor.java
kokosing: hm... it is just a wrapper of `Files.copy`. It just suppresses the exception.

findepi: But is very specific usage: to feed input to a process. You rather don't call it "copying".

PR #264, opened 2018-06-25
Remove deprecated parseTimestampInLocalTime

findepi:
There are usages in Presto, I will fix them as well.


findepi:
requires https://github.com/prestodb/presto/pull/11041 first

PR #267, opened 2018-10-18
Support escaping end of the line in convention tests

kokosing:
Support escaping end of the line in convention tests

### DIFF
--- a/tempto-core/src/main/java/io/prestodb/tempto/internal/convention/AnnotatedFileParser.java
+++ b/tempto-core/src/main/java/io/prestodb/tempto/internal/convention/AnnotatedFileParser.java
@@ -67,6 +67,7 @@ public class AnnotatedFileParser
     private static final String COMMENT_PREFIX = "---";
     private static final String SECTION_PREFIX = "--!";
     private static final String SECTION_NAME_KEY = "name";
+    private static final String NEW_LINE_LITERAL = "\\n";
 
     private static final Splitter.MapSplitter COMMENT_PROPERTIES_SPLITTER = Splitter.on(';')
             .omitEmptyStrings()
@@ -143,11 +144,11 @@ public class AnnotatedFileParser
      */
     private List<String> filterContent(List<String> contentLines)
     {
-        contentLines = contentLines.stream()
+        return contentLines.stream()
                 .filter(s -> !(isSpecialLine(s) || isBlank(s)))
                 .map(AnnotatedFileParser::unescapeLine)
+                .map(line -> line.replace(NEW_LINE_LITERAL, "\n"))
                 .collect(toList());
-        return contentLines;
     }
 
     private static String unescapeLine(String s)
--- a/tempto-core/src/test/groovy/io/prestodb/tempto/internal/convention/AnnotatedFileParserTest.groovy
+++ b/tempto-core/src/test/groovy/io/prestodb/tempto/internal/convention/AnnotatedFileParserTest.groovy
@@ -107,6 +107,21 @@ class AnnotatedFileParserTest
         sections[1].sectionName.get() == 'section2'
     }
 
+    def 'handles line end escapes'()
+    {
+        String fileContent = '--! name: section1\n' +
+                'line\\n1\n' + 
+                'line2\n' +
+                '--! name: section2'
+        def sections = fileParser.parseFile(toInputStream(fileContent))
+
+        expect:
+        sections.size() == 2
+        sections[0].sectionName.get() == 'section1'
+        sections[0].contentLines == ["line\n1", "line2"]
+        sections[1].sectionName.get() == 'section2'
+    }
+
     def parseOnlySection(String fileContent)
     {
         getOnlyElement(fileParser.parseFile(toInputStream(fileContent)))


tempto-core/src/main/java/io/prestodb/tempto/internal/convention/AnnotatedFileParser.java
```diff
@@ -143,11 +144,22 @@ private int findNextSectionIndex(List<String> lines)
...
+        ImmutableList.Builder<String> result = ImmutableList.builder();
+
+        for (int i = 0; i < contentLines.size(); i++) {
+            String line = contentLines.get(i);
+            if (isSpecialLine(line) || isBlank(line)) {
+                continue;
+            }
+
+            while (line.endsWith(LINE_ESCAPE)) {
+                checkState(i + 1 < contentLines.size(), "There no more lines, while the end of the line was escaped");
```
findepi:
```suggestion
                checkState(i + 1 < contentLines.size(), "There are no more lines, while the end of the line was escaped");
```

tempto-core/src/main/java/io/prestodb/tempto/internal/convention/AnnotatedFileParser.java
```diff
@@ -143,11 +144,22 @@ private int findNextSectionIndex(List<String> lines)
...
+                line = line.substring(0, line.length() - 1) + "\n" + contentLines.get(++i);
```
findepi: - `line.substring(0, line.length() - 1)` is hardly readable. i would use `replaceAll(".$", "")` instead
- don't increment `i` inline (i.e. don't treat `i++` or `++i` as an expression)
- this should be a `while` loop -- you shouldn't modify `for`'s loop index variablae
- this is quadratic

tempto-core/src/main/java/io/prestodb/tempto/internal/convention/AnnotatedFileParser.java
```diff
@@ -143,11 +144,22 @@ private int findNextSectionIndex(List<String> lines)
...
-        return contentLines;
+        ImmutableList.Builder<String> result = ImmutableList.builder();
+
+        for (int i = 0; i < contentLines.size(); i++) {
+            String line = contentLines.get(i);
+            if (isSpecialLine(line) || isBlank(line)) {
+                continue;
+            }
+
+            while (line.endsWith(LINE_ESCAPE)) {
```
findepi: - what if the line ends with `\\` or `\\\` ?
- instead of glueing the lines with `\<linefeed>`, can we use literal `\n` to represent a new line character in the data?

tempto-core/src/test/groovy/io/prestodb/tempto/internal/convention/AnnotatedFileParserTest.groovy
```diff
@@ -107,6 +107,20 @@ class AnnotatedFileParserTest
         sections[1].sectionName.get() == 'section2'
     }
 
+    def 'handles line end escapes'()
+    {
+        String fileContent = '--! name: section1\n' +
+                'line1\\\nline2\\\nline3\n' +
```
findepi: put each data input line as separate source file line

kokosing:
@findepi comments addressed.

tempto-core/src/main/java/io/prestodb/tempto/internal/convention/AnnotatedFileParser.java
```diff
@@ -143,11 +144,11 @@ private int findNextSectionIndex(List<String> lines)
      */
     private List<String> filterContent(List<String> contentLines)
     {
-        contentLines = contentLines.stream()
+        return contentLines.stream()
                 .filter(s -> !(isSpecialLine(s) || isBlank(s)))
+                .map(line -> line.replace(NEW_LINE_LITERAL, "\n"))
```
findepi: - this definitely needs to go _after_ `.map(AnnotatedFileParser::unescapeLine)`
- can `\n` in the file be escaped by yet another `\`?

tempto-core/src/test/groovy/io/prestodb/tempto/internal/convention/AnnotatedFileParserTest.groovy
```diff
@@ -107,6 +107,20 @@ class AnnotatedFileParserTest
         sections[1].sectionName.get() == 'section2'
     }
 
+    def 'handles line end escapes'()
+    {
+        String fileContent = '--! name: section1\n' +
+                'line\\n1\nline2\n' +
```
findepi:
```suggestion
                'line\\n1\n' + 
                'line2\n' +
```

tempto-core/src/main/java/io/prestodb/tempto/internal/convention/AnnotatedFileParser.java
kokosing: > can \n in the file be escaped by yet another \?

There is no such thing like escaping characters in data files.

findepi: so now it won't be possible to have literal `\n` (not `␊`) in the data file, right?

kokosing: right. It is like `\NULL` which stands for `null`.

PR #273, opened 2022-03-29
Update TestNG

sdruzkin:
This change updates TestNG from 6.10 to 6.14.3.

The new version has dedicated asserts for array, that's why I want to update TestNG in Presto https://github.com/prestodb/presto/pull/17521 to make tests more performant. 

I need to update TestNG in this project first because there was an API change between TestNG 6.10 and 6.14.3 in ITestNGMethod class. Due to the API change tests in https://github.com/prestodb/presto/pull/17521 fail.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -30,7 +30,7 @@ ext.versions = [
         guava        : '21.0',
         guice        : '4.0',
         log4j        : '1.2.17',
-        testng       : '6.10',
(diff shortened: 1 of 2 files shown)


sdruzkin:
@arhimondr Can you please take a look at this PR, and merge + release a new version if it looks good to you?

PR #279, opened 2024-11-27
upgraded_kafka_client_version_to_3.9.0

adkharat:
Existing kafka version = 0.11.0.2
Upgraded Kafka version = 3.9.0

With reference to prestodb/presto PR [Fix CVE-2022-34917 ](https://github.com/prestodb/presto/pull/24097)

Kafka 3.9.0 depricates direct Zookeeper interactions:

- ZkClient, 
- ZkUtils

Upgrading depricates below AdminUtils methods:

- AdminUtils.createTopic() 
- AdminUtils.topicExists()
- AdminUtils.deleteTopic()

1. Deprecated AdminUtils functions are replaced with **AdminClient** APIs
2. Deprecated Zookeeper dependency is replaced with Kafka's AdminClient API.
3. Encapsulated AdminClient logic in a method **KafkaAdminClient** to manage lifecycle and reduce boilerplate.
4. Operations such as topic creation (createTopic) and deletion (deleteTopic) now use AdminClient's **createTopics** and **deleteTopics**.

FYI:

kafka.utils.ZkUtils  was deprecated since 2.0.0.

2.0.0- > ZKUtils present
2.3.1 -> ZKUtils present
2.4.0 -> ZKUtils removed
2.8.2 -> ZKUtils not present (Version 2.8.2 has some vulnerability)
3.7.1 ->  ZKUtils not present (No vulnerability)
3.9.0 ->  ZKUtils not present (Latest version + No vulnerability)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -57,8 +57,7 @@ ext.versions = [
         cassandra    : '3.4.0',
         commons_cli  : '1.3.1',
         thrift       : '0.9.3',
-        kafka        : '0.11.0.2',
-        zkclient     : '0.10'
+        kafka        : '3.9.0'
 ]
 
 ext.libraries = [
@@ -102,8 +101,6 @@ ext.libraries = [
         commons_cli          : "commons-cli:commons-cli:${versions.commons_cli}",
         thrift               : "org.apache.thrift:libthrift:${versions.thrift}",
         kafka_clients        : "org.apache.kafka:kafka-clients:${versions.kafka}",
-        kafka                : "org.apache.kafka:kafka_2.12:${versions.kafka}",
-        zkclient             : "com.101tec:zkclient:${versions.zkclient}"
 ]
 
 ext.tempto_core = project(':tempto-core')
@@ -113,7 +110,7 @@ ext.tempto_runner = project(':tempto-runner')
 ext.tempto_ldap = project(':tempto-ldap')
 ext.tempto_kafka = project(':tempto-kafka')
 ext.expected_result_generator = project(':expected-result-generator')
-ext.tempto_version = '1.53'
+ext.tempto_version = '1.54'
 ext.tempto_group = "io.prestodb.tempto"
 ext.isReleaseVersion = !tempto_version.endsWith("SNAPSHOT")
 
--- a/tempto-examples/build.gradle
+++ b/tempto-examples/build.gradle
@@ -41,6 +41,7 @@ jar {
 
 shadowJar {
     version = ''
+    zip64 = true
 }
 
 build.dependsOn.add([shadowJar])
--- a/tempto-kafka/src/main/java/io/prestodb/tempto/fulfillment/table/kafka/KafkaTableManager.java
+++ b/tempto-kafka/src/main/java/io/prestodb/tempto/fulfillment/table/kafka/KafkaTableManager.java
@@ -24,47 +24,47 @@ import io.prestodb.tempto.fulfillment.table.TableManager;
 import io.prestodb.tempto.internal.fulfillment.table.TableName;
 import io.prestodb.tempto.query.QueryExecutor;
 import io.prestodb.tempto.query.QueryResult;
-import kafka.admin.AdminUtils;
-import kafka.admin.RackAwareMode;
-import kafka.utils.ZKStringSerializer$;
-import kafka.utils.ZkUtils;
-import org.I0Itec.zkclient.ZkClient;
-import org.I0Itec.zkclient.ZkConnection;
+import org.apache.kafka.clients.admin.AdminClient;
+import org.apache.kafka.clients.admin.AdminClientConfig;
+import org.apache.kafka.clients.admin.NewTopic;
+import org.apache.kafka.clients.admin.DeleteTopicsResult;
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.Producer;
 import org.apache.kafka.clients.producer.ProducerRecord;
+import org.slf4j.Logger;
 
 import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Singleton;
 
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.Properties;
+import java.util.concurrent.ExecutionException;
 import java.util.function.Consumer;
 
 import static java.lang.String.format;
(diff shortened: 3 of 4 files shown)


adkharat:
System configuration to test build
gradle -v                                                     
------------------------------------------------------------
Gradle 6.6
------------------------------------------------------------

Build time:   2020-08-10 22:06:19 UTC
Revision:     d119144684a0c301aea027b79857815659e431b9

Kotlin:       1.3.72
Groovy:       2.5.12
Ant:          Apache Ant(TM) version 1.10.8 compiled on May 10 2020
JVM:          1.8.0_422 (Azul Systems, Inc. 25.422-b05)
OS:           Mac OS X 14.4 x86_64

java -version 
openjdk version "1.8.0_422"
OpenJDK Runtime Environment (Zulu 8.80.0.17-CA-macosx) (build 1.8.0_422-b05)
OpenJDK 64-Bit Server VM (Zulu 8.80.0.17-CA-macosx) (build 25.422-b05, mixed mode)

adkharat:
@imjalpreet can you please review the changes.

tempto-examples/build.gradle
```diff
@@ -41,6 +41,7 @@ jar {
 
 shadowJar {
     version = ''
+    zip64 = true
```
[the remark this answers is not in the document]
adkharat: @ZacBlanco on gradle build, I am getting error 

```
Execution failed for task ':tempto-examples:shadowJar'.
> shadow.org.apache.tools.zip.Zip64RequiredException: archive contains more than 65535 entries.
```


image

Here, ShadowJar plugin is creating a JAR file that exceeds the limit of 65,535 entries. This might due to the ZIP format limitation that ShadowJar uses by default. [Link](https://github.com/GradleUp/shadow/issues/107)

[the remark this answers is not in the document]
ZacBlanco: Got it, thanks for the explanation. This is fine then I guess. We were probably close to the limit and the newer version pushed us over? It would be good to know the number of entries before/after this change, but I don't think it's worth trying to limit

ZacBlanco (commented):
Only other concern is the same with Jalpreet's on the presto PR. I think we should try to move to the latest available Kafka version

adkharat:
@ZacBlanco Latest Kafka version is 3.9.0. 
Pushed changes for 3.9.0

build.gradle
```diff
@@ -113,7 +111,7 @@ ext.tempto_runner = project(':tempto-runner')
 ext.tempto_ldap = project(':tempto-ldap')
 ext.tempto_kafka = project(':tempto-kafka')
 ext.expected_result_generator = project(':expected-result-generator')
-ext.tempto_version = '1.53'
+ext.tempto_version = '1.54'
```
ZacBlanco, line 114: This should only be upgraded and set to a non-SNAPSHOT version during the release process. Since we already release 1.53, this should be `1.54-SNAPSHOT`. During release we will update to 1.54, and then move to 1.55-SNAPSHOT afterwards.

So could you set this to 1.54-SNAPSHOT for now?

ZacBlanco (changes requested):
Just one comment, otherwise LGTM.

ZacBlanco (approved):

ZacBlanco:
I can't merge it. I don't have write access to this repository. @imjalpreet if you have time could you also review?

The jar is not automatically published. Someone with write access has to publish it to maven central. We also need to verify that the latest jar update doesn't break the presto-product-tests CI on the presto repository.

build.gradle
```diff
@@ -102,8 +101,7 @@ ext.libraries = [
         commons_cli          : "commons-cli:commons-cli:${versions.commons_cli}",
         thrift               : "org.apache.thrift:libthrift:${versions.thrift}",
         kafka_clients        : "org.apache.kafka:kafka-clients:${versions.kafka}",
-        kafka                : "org.apache.kafka:kafka_2.12:${versions.kafka}",
-        zkclient             : "com.101tec:zkclient:${versions.zkclient}"
+        kafka                : "org.apache.kafka:kafka_2.12:${versions.kafka}"
```
agrawalreetika, line 104: Curious, which class is getting used from this library?

tempto-kafka/src/main/java/io/prestodb/tempto/fulfillment/table/kafka/KafkaTableManager.java
```diff
@@ -95,6 +90,17 @@ public TableInstance<KafkaTableDefinition> createImmutable(KafkaTableDefinition
         return new KafkaTableInstance(createdTableName, tableDefinition);
     }
 
+    private void withKafkaAdminClient(Consumer<AdminClient> routine) {
+        Properties properties = new Properties();
+        properties.put(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, brokerHost + ":" + brokerPort);
+        AdminClient adminClient = AdminClient.create(properties);
+        try {
```
agrawalreetika, line 97: We can use `try-with-resources` instead for handling close of adminClient

tempto-kafka/src/main/java/io/prestodb/tempto/fulfillment/table/kafka/KafkaTableManager.java
```diff
@@ -95,6 +90,17 @@ public TableInstance<KafkaTableDefinition> createImmutable(KafkaTableDefinition
...
+            routine.accept(adminClient);
```
agrawalreetika, line 98: What would happen if there is an issue/exception while doing above?

agrawalreetika (changes requested):

build.gradle
agrawalreetika: Hi @adkharat 

I just checked I see below classes that you mentioned are coming from `kafka-clients-3.9.0.jar` nothing from these are from `org.apache.kafka:kafka_2.12:${versions.kafka}`

```
org/apache/kafka/clients/producer/KafkaProducer.class
org/apache/kafka/clients/admin/AdminClient.class
org/apache/kafka/common/errors/TopicExistsException.class
org/apache/kafka/common/serialization/ByteArraySerializer.class
```
So I am curious do we even need `"org.apache.kafka:kafka_2.12:${versions.kafka}"` in this upgrade?

ZacBlanco:
@adkharat I am closing this since we merged #288

PR #282, opened 2024-12-12
Add GitHub Action for Gradle build in Java CI

Dilli-Babu-Godari:
Added a GitHub Action workflow to run Gradle build for the prestodb/tempto repository. This action automates the CI process for Java projects and ensures consistency in builds for every push and pull request to the master branch.

The workflow includes:
- Using Ubuntu-latest as the build runner.
- Setting up JDK using actions/setup-java@v4 with Temurin distribution.
- Configuring gradle.properties for build settings.
- Executing the Gradle build process with ./gradlew build.

Also updated .gitignore to exclude:
- The .gradle/ directory while retaining gradle.properties.
- The build/ directory.

This enhancement streamlines the build process.

PR: https://github.com/Dilli-Babu-Godari/tempto/pull/2 
This PR serves as proof that the CI configuration functions correctly.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -1,0 +1,38 @@
+name: Java CI with Gradle
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+      matrix:
+        java-version:
+          - 8
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          show-progress: false
+          fetch-depth: 0
+
+      # Step to set up JDK
+      - name: Set up JDK
+        uses: actions/setup-java@v4
+        with:
+          java-version: ${{ matrix.java-version }}
+          distribution: 'temurin'
+          cache: gradle
+
+      # Step to set up gradle.properties
+      - name: Set up gradle.properties
+        run: |
+          echo "skipSigning=false" >> gradle.properties
+
+      # Step to run Gradle build
+      - name: Gradle Build
+        run: ./gradlew build


.gitignore
```diff
@@ -1,6 +1,6 @@
 # Build files
 .gradle/
-gradle.properties
```
ZacBlanco, line 3: I don't think this ignore rule should be changed? The CI should run without committing the `gradle.properties` file.

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,39 @@
...
+        with:
+          java-version: ${{ matrix.java-version }}
+          distribution: 'temurin'
+          cache: gradle
+
+      # Step to set up gradle.properties
+      - name: Set up gradle.properties
+        run: |
+          cp gradle.properties.example gradle.properties
+          grep -q '^skipSigning=' gradle.properties || echo "skipSigning=false" >> gradle.properties
```
ZacBlanco, line 35: I don't think you need the `cp` or `grep` here. I think it can just be as simple as:

```
echo "skipSigning=false" >> gradle.properties
```

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      # Step to run Gradle build
+      - name: Gradle Build
+        run: ./gradlew build
```
ZacBlanco, line 39: nit: can you add a newline here?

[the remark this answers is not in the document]
ZacBlanco: I think the `build` task includes testing. See https://stackoverflow.com/questions/44324060/what-does-gradle-build-task-include-exactly

ZacBlanco (changes requested):

.gitignore
Dilli-Babu-Godari: > I don't think this ignore rule should be changed? The CI should run without committing the `gradle.properties` file.

reverted back the change in this file.

.github/workflows/gradle.yml
Dilli-Babu-Godari: > I don't think you need the `cp` or `grep` here. I think it can just be as simple as:
> 
> ```
> echo "skipSigning=false" >> gradle.properties
> ```

made the suggested changes.

.github/workflows/gradle.yml
Dilli-Babu-Godari: > nit: can you add a newline here?

I have added a new line. Please take a look at the changes now.

ZacBlanco (approved):

.github/workflows/gradle.yml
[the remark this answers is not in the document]
tdcmeehan: I think you're right.  But at the same time, I don't see any logging around tests being run in the example PR, do you?  https://github.com/Dilli-Babu-Godari/tempto/actions/runs/12320197877/job/34388934218?pr=2

I would expect at least a summary of the tests executed.  @Dilli-Babu-Godari can you add another example with a deliberately failing test case so we can see how it looks?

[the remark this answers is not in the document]
Dilli-Babu-Godari: > I think you're right. But at the same time, I don't see any logging around tests being run in the example PR, do you? https://github.com/Dilli-Babu-Godari/tempto/actions/runs/12320197877/job/34388934218?pr=2
> 
> I would expect at least a summary of the tests executed. @Dilli-Babu-Godari can you add another example with a deliberately failing test case so we can see how it looks?

I've made the changes as requested. I added a deliberately failing test case, and you can see the result in the build logs. The test case has failed, and the logs reflect this failure. I’ve attached the results for reference.

Screenshot 2024-12-13 at 11 45 40 PM

tdcmeehan (approved):

PR #292, opened 2025-02-24
Fix gpg signing

unidevel:
useGpgCmd seems never work on both my machine and the action env, not sure why.

So replace it with `useInMemoryPgpKeys` without using the gpg command, also added 2 checkings to avoid errors when running build/test in dev env(without the gradle.properties, or not set some of the signing related properties)

Tested with https://github.com/unix280/tempto/actions/runs/13507065551

Signature can be found in my test repo https://m2.unidevel.cn/#/releases/io/prestodb/tempto/tempto-core/1.54

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,6 +12,9 @@
  * limitations under the License.
  */
 
+import java.nio.file.Files
+import java.nio.file.Paths
+
 plugins {
     id "com.github.hierynomus.license" version "0.16.1"
     id 'signing'
@@ -151,8 +154,18 @@ configure(subprojects - expected_result_generator - tempto_examples) {
     }
 
     signing {
+        def skipSigning = project.findProperty('skipSigning')?.toBoolean() != false
+        if (!skipSigning) {
+            def secretKeyRingFilePath = project.findProperty("signing.secretKeyRingFile")?.with { it.isEmpty()? null : it }
+            def signingKeyId = project.findProperty("signing.keyId")?.with { it.isEmpty() ? null : it }
+            def signingPassword = project.findProperty("signing.password")?.with { it.isEmpty()? null : it }
+            def keyFile = secretKeyRingFilePath ? Paths.get(secretKeyRingFilePath) : null
+            if (keyFile && Files.exists(keyFile) && !Files.isDirectory(keyFile)) {
+                def secretKey = new String(Files.readAllBytes(keyFile), "UTF-8")
+                useInMemoryPgpKeys(signingKeyId, secretKey, signingPassword)
+            }
+        }
         sign publishing.publications
-        useGpgCmd() //Use system provide GPG
     }
 
     // publishing to sonatype


build.gradle
```diff
@@ -151,8 +154,18 @@ configure(subprojects - expected_result_generator - tempto_examples) {
...
     signing {
+        def skipSigning = project.findProperty('skipSigning')?.toBoolean() != false
+        if (!skipSigning) {
+            def secretKeyRingFilePath = project.findProperty("signing.secretKeyRingFile")?.with { it.isEmpty()? null : it }
+            def signingKeyId = project.findProperty("signing.keyId")?.with { it.isEmpty() ? null : it }
+            def signingPassword = project.findProperty("signing.password")?.with { it.isEmpty()? null : it }
+            def keyFile = secretKeyRingFilePath ? Paths.get(secretKeyRingFilePath) : null
+            if (keyFile && Files.exists(keyFile) && !Files.isDirectory(keyFile)) {
+                def secretKey = new String(Files.readAllBytes(keyFile), "UTF-8")
+                useInMemoryPgpKeys(signingKeyId, secretKey, signingPassword)
```
unidevel, line 165: use null value to make the parameter optional, e.g. user set the `signing.keyId=`, if use empty string here, it will report error like `secret key not found`, but if use the null value here, it will use the first secret key in the secret file

build.gradle
```diff
@@ -151,8 +154,18 @@ configure(subprojects - expected_result_generator - tempto_examples) {
     }
 
     signing {
+        def skipSigning = project.findProperty('skipSigning')?.toBoolean() != false
```
ZacBlanco, line 157: Can we leave out the comparison?

```suggestion
        def skipSigning = project.findProperty('skipSigning')?.toBoolean()
```

build.gradle
ZacBlanco: I am assuming that the signing failed previously because we need to import the key into the actions runner's GPG keyring. It failed because [no key was found on the keyring](https://github.com/prestodb/tempto/actions/runs/13501165970/job/37719706150#step:14:56). Could we try the import as a setup step in the action instead of calling the `useInMemoryPgpKeys`? I think it just needs a `gpg --import keyFile.gpg` or something like that.

Have you tried using [nektos/act](https://github.com/nektos/act) for testing?

unidevel:
> I am assuming that the signing failed previously because we need to import the key into the actions runner's GPG keyring. It failed because [no key was found on the keyring](https://github.com/prestodb/tempto/actions/runs/13501165970/job/37719706150#step:14:56). Could we try the import as a setup step in the action instead of calling the useInMemoryPgpKeys? I think it just needs a gpg --import keyFile.gpg or something like that.

The problem is that `gradle sign` will use `gpg --batch` which can't accept any passpharse, unless it is cached before. 

That's why I'm using `useInMemoryPgpKeys` which has the minimal change comparing with `useGpgCmd`

> Have you tried using [nektos/act](https://github.com/nektos/act) for testing?

Good to know this, will try it in future, thanks

ZacBlanco (approved):
Thanks for the changes. After taking another look at the signing plugin docs I think this looks good.

unidevel:
@ZacBlanco Thanks for your comments, I found another way to cache the passphase which is simpler than this, so I'll close this one.

PR #293, opened 2025-02-24
Fix gpg signing with the passphrase cache

unidevel:
This is another way to fix the gpg signing issue with the passphrase cache by signing a dummy file.

Tested with https://github.com/unix280/tempto/actions/runs/13509698265/job/37747324801

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,6 +12,8 @@ jobs:
     permissions:
       contents: write
       packages: write
+    env:
+      GPG_KEY_FILE: ${{ github.workspace }}/secret-key.gpg
 
     steps:
       - name: Check branch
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -96,15 +98,19 @@ jobs:
 
       - name: Set up GPG key
         run: |
-          echo "${{ secrets.GPG_SECRET }}" > ${{ github.workspace }}/secret-key.gpg
-          chmod 600 ${{ github.workspace }}/secret-key.gpg
+          echo "${{ secrets.GPG_SECRET }}" > ${{ env.GPG_KEY_FILE }}
+          chmod 600 ${{ env.GPG_KEY_FILE }}
+          gpg --import --batch ${{ env.GPG_KEY_FILE }}
+          touch /tmp/dummy.txt && gpg --batch --yes --pinentry-mode loopback --passphrase "${{ secrets.GPG_PASSPHRASE }}" --sign /tmp/dummy.txt
```
unidevel, line 104: Cached the passphrase here

ZacBlanco (approved):

unidevel:
> It looks like you actually ran the release action. Can you verify if the GPG signing checks at sonatype central were able to pass?

@ZacBlanco I'm using a test gpg key, which not send to the public key servers. The GPG keys used in action are  set as secrets, I can not verify it locally.

PR #294, opened 2025-02-26
Add more pom information to artifacts

aaneja:
to meet Maven central requirements
Also, add back ability to skip signing by using `-PskipSigning=true` during builds


ZacBlanco (approved):

unidevel (approved):

unidevel:
Thanks @aaneja , let's try to release again.

